PR #20, opened 2020-04-25
Format code using 'cargo fmt'

Atul9:


mrDIMAS:
Thanks for PR, but I applied rustfmt by myself since there were lots of conflicts.

PR #36, opened 2020-10-09
Add script to install latest rg3g components (unix)

alekspickle:
I guess its doable to make .bat for windows users.


mrDIMAS (changes requested):

mrDIMAS:
I'm closing this because I've moved dependencies in workspace so there is no need to download them separately.

PR #37, opened 2020-10-10
Add serde integration

jwinnie:
This change adds a flag called `serde_integration` that enables `#[derive(Serialize, Deserialize)]` on `QualitySettings` and enables Serde support in Glutin.

(please squash when merging because I made a bunch of commits experimenting with random stuff that shouldn't make it into the history)


src/renderer/mod.rs
```diff
@@ -336,7 +339,7 @@ pub struct Renderer {
     quad: SurfaceSharedData,
     frame_size: (u32, u32),
     ambient_color: Color,
-    quality_settings: QualitySettings,
+    pub quality_settings: QualitySettings,
```
mrDIMAS, line 342: This should be private, check `set_quality_settings` of Renderer, it doing a lot more than just setting a field. If it would be exposed as public then setting new value to it wouldn't have effect.

jwinnie: Ah okay. I didn't know that method existed.

jwinnie: @mrDIMAS resolved. PR ready to be merged

PR #48, opened 2020-11-01
improved sound example

itmuckel:
* Get resource references before event_loop starts
* Align sounds with animation

### DIFF
--- a/examples/sound.rs
+++ b/examples/sound.rs
@@ -61,12 +61,24 @@ fn main() {
     let mut elapsed_time = 0.0;
 
     // We'll use four footstep sounds to randomize sound and make it more natural.
-    let footsteps = [
+    let footstep_paths = [
         "examples/data/sounds/FootStep_shoe_stone_step1.wav",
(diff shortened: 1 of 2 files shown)


examples/sound.rs
```diff
@@ -61,13 +61,26 @@ fn main() {
...
+        .iter()
+        .map(|&path| {
+            rg3d::futures::executor::block_on(
+                game.engine
+                    .resource_manager
+                    .request_sound_buffer(path, false),
+            )
+            .unwrap()
+        })
+        .collect::<Vec<_>>();
```
mrDIMAS, line 82: The problem with this code is that you waiting on each buffer separately, please consider to change it to this:

```rust
let footstep_buffers = rg3d::futures::executor::block_on(rg3d::futures::future::join_all(
	footstep_paths.iter().map(|&path| {
		game.engine
			.resource_manager
			.request_sound_buffer(path, false)
	}),
))
.into_iter()
.map(|r| r.unwrap())
.collect::<Vec<_>>();
```

This code loads all sound buffers in parallel.

PR #63, opened 2020-11-22
added environment mapping

itmuckel:


src/renderer/gbuffer.rs
```diff
@@ -329,6 +351,24 @@ impl GBuffer {
                                     texture: batch.lightmap_texture.clone(),
                                 },
                             ),
+                            (
+                                self.shader.camera_position,
+                                UniformValue::Vector3(camera.local_transform().position()),
```
mrDIMAS, line 356: Should be `camera.global_position()`, otherwise if camera is attached to some other node you'll get position relative to that node, but you need world-space position.

mrDIMAS (changes requested):

PR #67, opened 2020-12-29
Improve READMEs, fix typos

martin-t:
Let me know if you disagree with anything.

The logo might look better if it was a bit smaller but i haven't looked at how to edit the svg.

I changed the heading to "Rust Game 3D" because as i said on discord, it's not immediately obvious what RG3D stands for and i think this way it's more memorable.

There's also a way to center the heading/logo if you think it would look better.

What do you think about including the number of comments https://tokei.rs/b1/github/mrDIMAS/rg3d?category=comments among the badges?

I changed the bit about ECS/OOP because i think a lot of people imagine inheritance when they hear OOP and it might not evoke positive emotions, especially in the rust crowd ;)


toyboot4e:
Also, I wonder if we substitute WARNING to NOTE in sub crates' README.md so that they have positive nuance :)

martin-t:
Definitely, i didn't notice the subcrates has separate readmes. I skimmed them and made some more fixes.

README.md
```diff
@@ -1,13 +1,14 @@
+[![RG3D](pics/logo2.svg)](https://rg3d.rs/)
+
+# Rust Game 3D
```
mrDIMAS, line 3: It should be **R**ust **G**ame engine **3D**, I think it is less confusing than Rust Game - which looks like it is game, not engine.

mrDIMAS (changes requested):
Could you please change that one comment? The rest looks good, it is a shame that I missed so many articles :smile:

README.md
mrDIMAS: Actually, I just realized that I can fix it by myself after merge

PR #89, opened 2021-03-27
Add Context Menu

### LINKED ISSUE #81: Implement context menus
rg3d-ui still does not have context menus, they're very important for better user experience. Possible solution:
1) Add `menu: Handle<UINode<M, C>>` field to Widget 
2) In `UserInterface::poll_message` handle MouseDown events: on each click check for any node with not empty `menu` field up on visual tree starting from picked node.
3) If any node with not empty `menu` was found, send `PopupMessage::placement` with position of mouse cursor (`Placement::Cursor`) to `menu` and also send `PopupMessage::open`
Here we assume that `menu` is an instance of Popup widget, other types of widgets are not supported here. By the type I mean "any entity that can handle `PopupMessage`s"

MinusGix:
This PR adds support for context menus (closes #81), essentially popups that are shown on right click. Example usage:
```rust
WidgetBuilder::new().with_context_menu(
    PopupBuilder::new(WidgetBuilder::new())
        .with_content(
            BorderBuilder::new(
                WidgetBuilder::new().with_child(
                    TextBuilder::new(WidgetBuilder::new())
                        .with_text("Testing context menu!")
                        .build(ctx),
                ),
            )
            .with_stroke_thickness(Thickness::uniform(1.0))
            .build(ctx),
        )
        .stays_open(true)
        .build(ctx)
    )
    .build(),
```

Though, I think this would be more useful if `Menu` could be used for this, since often context menus are going to look and act really similar (one appearing at the button, the other at cursor) with both being list of options. Perhaps make `Menu` handle `PopupMessage`s as well?

### DIFF
--- a/rg3d-ui/src/lib.rs
+++ b/rg3d-ui/src/lib.rs
@@ -13,6 +13,8 @@ extern crate lazy_static;
 extern crate fontdue;
 extern crate sysinfo;
 
+use message::PopupMessage;
+use popup::Placement;
 pub use rg3d_core as core;
 
 pub mod border;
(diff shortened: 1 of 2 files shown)


mrDIMAS:
Using of menu here would be too restricting, because context menu basically may have anything inside of it. For example Firefox has context menu with complex layout - it has first row with buttons and rest of space is taken by stack panel with menu items. What we should really do, is to make MenuItem working without Menu - right now MenuItem relies on the presense of Menu, which handle activation and deactivation (which in turn hides menu chain).

PR #91, opened 2021-03-28
Use Options for Handles Somewhat More

MinusGix:
This PR does not do anything too major. Mainly changing a few functions to returning `Option<Handle<T>>` rather than using `None`, and making some code within functions use `Option`s rather than `Handle::None`.  
(Essentially trying to split apart the big change of using `Option` with `Handle`s into smaller prs which are more manageable, and allow for better comments on progress)

# Commit #1  
  
## Return Changes  
- `Pool::handle_from_index` returns `Option<Handle<T>>` rather than `Handle<T>`.  
    - Note: StationIapetus uses this once. (Would make a PR for it, but see end)
- `NavMeshContainer::handle_from_index` See above.  
- `Graph::handle_from_index` See above.  
  
## Extra Methods  
- `Pool::at_pair(&self, n: usize) -> Option<(Handle<T>, &T)>` Returns the value and the handle to it. This avoids extra unneeded checks. (Ex: `rg3d-sound/src/context.rs` line 220)  
- `Pool::at_pair_mut(&mut self, n: usize) -> Option<(Handle<T>, &mut T)>`  
  
# Commit #2
This commit was (mostly) trying to only change some minor handling within functions. Replacing usage of `Handle::NONE` with `Option`.  
## Return changes
- `find_tree` (internal function) returns `Option<Handle<T>>` rather than `Handle<T>`  
- `unbind_by_body` returns `Option<Handle<T>>` rather than `Handle<T>`. This is public. (Grepped for it in my project, rusty-editor, rg3d-tutorials, StationIapetus, rusty-shooter and got no results for usage. Though, rusty-editor has its own version which is not changed.)
  
  
# Non-Added Methods
This section is for methods that were thought of but weren't actually added. Essentially to give an idea of what could be useful/nice to have but I didn't deem worthy enough to add at the moment.  
- `Pool::unwrap_valid_handle(&self, handle: Option<Handle<T>>) -> Option<Handle<T>>` If it is `None`, return `None`. If it is Invalid, return `None`. If it is Valid, return it. Allows:  
```rust
let handle: Option<Handle<T>> = pool.handle_from_index(i);
if let Some(handle) = pool.unwrap_valid_handle(handle) {
    //
}
```
(Ex: `rg3d-ui/src/lib.rs` line 1708)  
Trivial implementation: `handle.filter(|h| self.is_valid_handle(*h))`  
  
  
# Testing
Tested changes with:
 - `cargo test --release` in rg3d 
 - My own small project (No changes needed)
 - rusty-editor (No changes needed)
 - rg3d-tutorials (No changes needed)
 - StationIapetus: `src/level/mod.rs` line #688 uses `handle_from_index`. This should be `scene.navmeshes.handle_from_index().unwrap_or(Handle::None)` or use `.expect` if it should certainly exist. (For a bigger change to rg3d like a future PR might bring, I'd make a PR on Iapetus too, but this one is small)  
 - rusty-shooter (No changes needed)

### DIFF
--- a/rg3d-core/src/pool.rs
+++ b/rg3d-core/src/pool.rs
@@ -811,13 +811,42 @@ impl<T> Pool<T> {
 
     #[inline]
     #[must_use]
-    pub fn handle_from_index(&self, n: usize) -> Handle<T> {
+    pub fn at_pair(&self, n: usize) -> Option<(Handle<T>, &T)> {
         if let Some(record) = self.records.get(n) {
             if record.generation != INVALID_GENERATION {
-                return Handle::new(n as u32, record.generation);
+                return record
+                    .payload
+                    .as_ref()
+                    .map(|p| (Handle::new(n as u32, record.generation), p));
             }
         }
-        Handle::NONE
+        None
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn at_pair_mut(&mut self, n: usize) -> Option<(Handle<T>, &mut T)> {
+        if let Some(record) = self.records.get_mut(n) {
+            if record.generation != INVALID_GENERATION {
+                let generation = record.generation;
+                return record
+                    .payload
+                    .as_mut()
+                    .map(|p| (Handle::new(n as u32, generation), p));
+            }
+        }
+        None
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn handle_from_index(&self, n: usize) -> Option<Handle<T>> {
+        if let Some(record) = self.records.get(n) {
+            if record.generation != INVALID_GENERATION {
+                return Some(Handle::new(n as u32, record.generation));
+            }
+        }
+        None
     }
 
     /// Returns exact amount of "alive" objects in pool.
@@ -976,7 +1005,7 @@ impl<T> Pool<T> {
         self.records.as_ptr()
     }
 
-    pub fn handle_of(&self, ptr: &T) -> Handle<T> {
+    pub fn handle_of(&self, ptr: &T) -> Option<Handle<T>> {
         let begin = self.begin() as usize;
         let end = self.end() as usize;
         let val = ptr as *const T as usize;
@@ -987,7 +1016,7 @@ impl<T> Pool<T> {
                 return self.handle_from_index(record_location / record_size);
             }
         }
-        Handle::NONE
+        None
     }
 }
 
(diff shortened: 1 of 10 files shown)


rg3d-core/src/pool.rs
```diff
@@ -811,13 +811,25 @@ impl<T> Pool<T> {
 
     #[inline]
     #[must_use]
-    pub fn handle_from_index(&self, n: usize) -> Handle<T> {
+    pub fn at_pair(&self, n: usize) -> Option<(Handle<T>, &T)> {
+        self.handle_from_index(n).zip(self.at(n))
```
mrDIMAS, line 815: I think it does double validity checks:

1) `handle_from_index` checks if index and generation are both valid
2) `at` checks if index is valid

I'm not sure if compiler smart enough to eliminate such duplicated checks, so maybe it is worth to combine internals of `handle_from_index` + `at` to make more efficient version? I'm concerned about this method because it is used for iterators and **may** slowdown iterators. Possible replacement:

```rust
 pub fn at_pair(&self, n: usize) -> Option<(Handle<T>, &T)> {
	if let Some(record) = self.records.get(n) {
		if record.generation != INVALID_GENERATION {
			if let Some(payload) = record.payload.as_ref() {
				return Some((Handle::new(n as u32, record.generation), payload));
			}
		}
	}
	None
}
```

rg3d-core/src/pool.rs
```diff
@@ -1052,16 +1064,14 @@ impl<'a, T> Iterator for PoolPairIterator<'a, T> {
...
         loop {
-            match self.pool.records.get(self.current) {
-                Some(record) => {
-                    if let Some(payload) = &record.payload {
-                        let handle = self.pool.handle_from_index(self.current);
-                        self.current += 1;
-                        return Some((handle, payload));
-                    }
+            if self.pool.records.get(self.current).is_some() {
+                if let Some(pair) = self.pool.at_pair(self.current) {
```
mrDIMAS, line 1068: I think iterator has become less efficient with this, because of multiple checks of same condition:

1) `.get(self.current).is_some()` checks if index is valid
2) Current implementation of `at_pair` checks index two times

However, existing implementation of iterator does checks two times - it does same checks in `handle_from_index` at line 1058. At this point we've passed all checks already and maybe it is worth to just set `handle = Handle { generation: record.generation, index: self.current as u32 };`

mrDIMAS (changes requested):
Everything else looks fine, I'm concerned about three (`at_pair`, `at_pair_mut`, `PoolPairIterator::next()`) methods - they may be less efficient now :thinking: . Could you please "unroll" implementations of these and remove redundant checks?

PR #110, opened 2021-05-26
Adding Wasm CI

Voodlaz:


Voodlaz:
Should we use wasm-pack or test by some other way? What do you think?

.github/workflows/ci.yml
```diff
@@ -37,7 +37,26 @@ jobs:
...
       run: cargo test --verbose
-      
+
+  wasm:
+    name: Wasm CI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Dependencies
```
mrDIMAS, line 47: Not sure that this section is correct, according to this https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/continuous-integration.html all it needs to have is this:

```
- name: Install
        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
```

There is no need to install libraries because wasm target do not use them.

mrDIMAS (changes requested):

.github/workflows/ci.yml
Voodlaz: Ok

Voodlaz: I did it

mrDIMAS:
Ok, I hope it works, merging. Thanks.

PR #112, opened 2021-05-26
Fix CI

Voodlaz:


gideongrinberg:
Thanks @Voodlaz, I completely forgot to do that.

Issue #114, opened 2021-05-29
Draw skybox in a single draw call


mrDIMAS:
Currently, skybox is rendered as 6 separate planes with separate textures, but it could be drawn in a single draw call with cube map texture. Skybox is rendered in `renderer/deferred_light_renderer.rs`. To do this task, a set of sky box textures must be converted into a cube map (only on renderer side), and separate shader for rendering should be added. In the shader you could use position of vertex in world space to fetch cube map. As a mesh for skybox you could create cube geometry and flip it using -1 scale.

vigdail:
Hello, I would like to work on this issue, but I have a few questions:
1. At what stage we need to convert separate faces into a cubemap? Should there be something like `SkyBoxBuilder` to which we provide faces and then it builds the actual `SkyBox` with only a cubemap? Or we need to build the cubemap on demand before rendering?
2. How to handle the situation when some faces have not been provided? Missing parts should be filled with zeroes in `bytes` parameter of `Texture::from_bytes` method, right?

mrDIMAS:
Hi! The solution I've proposed is actually bad, it should be done as you wrote - make a SkyBoxBuilder which will construct a single cube map Texture. However there is one caveat - serialization. Currently, a texture for each side serialized separately, but when you'll convert a bunch of textures in a single one you'll lose connection with their respective source files. So to fix that, you should serialize textures for sides of skybox as usual (as it done in Visit trait implementation on SkyBox), but on load the cube map should be recreated. The point where cube map should be recreated is not so obvious - it can't be done directly in Visit trait implementation because in case of standalone scenes textures aren't pointing to actually loaded texture - they're just dummies with source file name inside - they'll be mapped to actual loaded resource on separate "resolving" stage. So the point to re-create cube map is `Scene::resolve` - what you need to do is just iterate over every camera and call your `SkyBox::update_cubemap` there. Also cube map should not be serialized, it will be "procedural" texture which is valid only for a single "session". At the renderer side you should use the cube map for rendering, instead of separate textures. Also there is already a mesh for the skybox so no need to create new one. As for missing textures - I think filling them with black (all zeros) is good.

PR #123, opened 2021-06-16
Skybox cubemap

vigdail:
There is an implementation of skybox with cubemap (#114) . Tbh, I don't like what `SkyBox::create_cubemap()` function looks like. Maybe your review will help me to find out the way to improve it.

### DIFF
--- a/examples/shared/mod.rs
+++ b/examples/shared/mod.rs
@@ -85,8 +85,8 @@ pub async fn create_camera(
 
     // Set S and T coordinate wrap mode, ClampToEdge will remove any possible seams on edges
     // of the skybox.
-    for skybox_texture in skybox.cubemap() {
-        let mut data = skybox_texture.data_ref();
+    if let Some(cubemap) = skybox.cubemap() {
+        let mut data = cubemap.data_ref();
         data.set_s_wrap_mode(TextureWrapMode::ClampToEdge);
         data.set_t_wrap_mode(TextureWrapMode::ClampToEdge);
     }
--- a/examples/shared/mod.rs
+++ b/examples/shared/mod.rs
@@ -30,7 +30,7 @@ use rg3d::{
     resource::texture::TextureWrapMode,
     scene::{
         base::BaseBuilder,
-        camera::{CameraBuilder, SkyBox},
+        camera::{CameraBuilder, SkyBoxBuilder},
         graph::Graph,
         node::Node,
         transform::TransformBuilder,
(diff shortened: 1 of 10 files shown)


examples/shared/mod.rs
```diff
@@ -72,18 +72,20 @@ pub async fn create_camera(
...
         bottom: Some(bottom.unwrap()),
-    };
+    }
+    .build()
+    .unwrap();
 
     // Set S and T coordinate wrap mode, ClampToEdge will remove any possible seams on edges
     // of the skybox.
-    for skybox_texture in skybox.textures().iter().filter_map(|t| t.clone()) {
+    for skybox_texture in skybox.cubemap() {
```
mrDIMAS, line 88: Looks like you iterating over Option here, maybe `if let Some(cubemap) = skybox.cubemap() {` will be better here?

src/renderer/shaders/skybox_vs.glsl
```diff
@@ -0,0 +1,15 @@
...
+
+uniform mat4 worldViewProjection;
+
+out vec3 texCoord;
+
+void main()
+{
+    texCoord = vertexPosition;
+    vec4 pos = worldViewProjection * vec4(vertexPosition, 1.0);
+    gl_Position = pos.xyww;
```
mrDIMAS, line 14: Just curious, what `.xyww` does here? Is it some trick I'm not aware of?

mrDIMAS (approved):
Thanks for the PR. There are few small notes, they're not critical and I'll approve anyway, but if you have some time to do a quick fix that would be awesome.

As for your question about `create_cubemap` - I could not imagine any other solution, your's should work fine.

examples/shared/mod.rs
vigdail: Thanks for the review. You're right `if let `would be more convenient here.

src/renderer/shaders/skybox_vs.glsl
vigdail: Oh, it makes `pos.z` equal to 1 after perspective division, which places vertex on the far plane of clip space, so that the skybox is always rendered behind other objects. But in our situation it's unnecessary, since we render the sky first. I removed it to avoid confusion.

vigdail:
I also removed UV attribute from skybox mesh, we don't need it to draw cubemap.

mrDIMAS:
CI seems to fail, but this is ok, I'll merge and fix that. Again, thanks for the PR!

PR #128, opened 2021-06-22
WASM start game by user interaction to enable sound

beckend:
Addresses: https://github.com/rg3dengine/rg3d/issues/99

Chrome needs user interaction to enable sound, added an animated/transition button to launch the demo for the heck of it.

start-game1
start-game2

### DIFF
--- a/examples/wasm/index.html
+++ b/examples/wasm/index.html
@@ -1,14 +1,20 @@
 <!DOCTYPE html>
 <html>
   <head>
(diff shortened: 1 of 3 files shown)


mrDIMAS:
Awesome, thanks for the PR! I guess every browser game use the same trick with explicit focusing? This issue is very annoying actually :thinking:

beckend:
I only tested on latest firefox and chromium which works fine.

PR #139, opened 2021-07-18
Add a mask_char to TextBox and FormattedText

VictorKoenders:
This adds a `with_mask_char` and `with_no_mask_char` to `TextBoxBuilder` and `FormattedText`.
Internally this changes the rendering of `FormattedText` to use a list of `mask_char` instead of `self.text` if the `mask_char` is set.
This does mean that multiline functionality does not work, as `code` is never `\n`, `\r` or a space, which is probably acceptable for password input fields.

image

### DIFF
--- a/rg3d-ui/src/formatted_text.rs
+++ b/rg3d-ui/src/formatted_text.rs
@@ -88,6 +88,7 @@ pub struct FormattedText {
     brush: Brush,
     constraint: Vector2<f32>,
     wrap: WrapMode,
+    mask_char: Option<char>,
 }
 
 #[derive(Copy, Clone, Debug)]
@@ -218,12 +219,21 @@ impl FormattedText {
             return Vector2::default();
         };
 
+        let masked_text: Vec<u32>;
+        let text: &Vec<u32> = if let Some(mask_char) = self.mask_char {
+            masked_text = (0..self.text.len()).map(|_| mask_char as u32).collect();
+            &masked_text
+        } else {
+            &self.text
+        };
+
+
         // Split on lines.
         let mut total_height = 0.0;
         let mut current_line = TextLine::new();
         let mut word: Option<Word> = None;
         self.lines.clear();
-        for (i, code) in self.text.iter().enumerate() {
+        for (i, code) in text.iter().enumerate() {
             let advance = match font.glyph(*code) {
                 Some(glyph) => glyph.advance,
                 None => font.height(),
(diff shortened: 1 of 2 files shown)


rg3d-ui/src/text_box.rs
```diff
@@ -906,6 +908,16 @@ impl<M: MessageData, C: Control<M, C>> TextBoxBuilder<M, C> {
         self
     }
 
+    pub fn with_mask_char(mut self, mask_char: char) -> Self {
```
mrDIMAS, line 911: I think one method that accepts `Option<char>` should be enough. `with_no_mask_char` is a bit confusing because you have no mask char by default.

rg3d-ui/src/formatted_text.rs
```diff
@@ -498,6 +510,16 @@ impl FormattedTextBuilder {
         self
     }
 
+    pub fn with_mask_char(mut self, mask_char: char) -> Self {
```
mrDIMAS, line 513: I think one method that accepts `Option<char>` should be enough. `with_no_mask_char` is a bit confusing because you have no mask char by default.

rg3d-ui/src/text_box.rs
```diff
@@ -917,21 +929,23 @@ impl<M: MessageData, C: Control<M, C>> TextBoxBuilder<M, C> {
...
         }
 
+        let mut formatted_text_builder = FormattedTextBuilder::new()
+                    .with_text(self.text)
+                    .with_font(self.font.unwrap_or_else(|| crate::DEFAULT_FONT.clone()))
+                    .with_horizontal_alignment(self.horizontal_alignment)
+                    .with_vertical_alignment(self.vertical_alignment)
+                    .with_wrap(self.wrap);
+        if let Some(mask_char) = self.mask_char {
+            formatted_text_builder = formatted_text_builder.with_mask_char(mask_char);
```
mrDIMAS, line 939: If `with_mask_char` would accept `Option<char>` then you could simply do 

```rust
...
    .with_mask_char(self.mask_char)
```

mrDIMAS (changes requested):
Few small comments, everything else looks good. Thanks for the PR!

mrDIMAS (approved):

VictorKoenders:
Ah I was confused with another crate that had `.with_` for setting values and `.with_no_` for resetting values. Updated!

PR #154, opened 2021-08-16
Allow creating SoundEngine without a sound device

### LINKED ISSUE #152: Support custom backends/devices in rg3d-sound
Currently the rg3d-sound library combines both sound processing and sound output. It would be nice to be able to use a different library for sound output (e.g. output to file, use SDL2 or CPAL, use dummy output for benchmarking etc.). Currently this is prevented by both mod device and context::State::render+SoundContext::SAMPLES_PER_CHANNEL being private.

The easiest way would probably be adding a couple of methods to SoundEngine:
```
pub fn render(buf: &mut [(f32, f32)])
pub fn required_buffer_len() -> u32
```

OlegOAndreev:
Fixes #152. Added an example of using the new methods to output a wav file.

### DIFF
--- a/rg3d-sound/examples/write_wav.rs
+++ b/rg3d-sound/examples/write_wav.rs
@@ -1,0 +1,64 @@
+use rg3d_sound::{
+    buffer::{DataSource, SoundBufferResource},
+    context::SoundContext,
+    pool::Handle,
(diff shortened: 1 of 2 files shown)


mrDIMAS (approved):
Looks good for me, thanks for the PR!

PR #157, opened 2021-08-21
fix allignment missmatch on alloc-dealloc

kazuk:
add MemoryLayout for new type of std::alloc::Layout

  with dummy Visitor implementation

add memory_layout field to VertexBuffer

  holds base Vec's memory layout

add fn VertexBuffer::replace_data
  replace data on VertexBuffer
with
  manage memory_layout information,
  ManuallyDrop for old data Vec with memory_layout

add fn VertexBuffer::extend_data_from_slice

  safe wrapper for Vec::extend_from_slice

add impl Drop for VertexBuffer

  alignment safe cleanup for VertexBuffer : using replace_data

### DIFF
--- a/src/scene/mesh/buffer.rs
+++ b/src/scene/mesh/buffer.rs
@@ -19,6 +19,7 @@ use crate::{
 };
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
+use std::mem::ManuallyDrop;
 use std::ops::{Deref, DerefMut, Index, IndexMut};
 use std::{marker::PhantomData, mem::MaybeUninit};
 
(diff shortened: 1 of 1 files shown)


mrDIMAS (changes requested):
I'm curious if such juggling with memory buffers is needed at all. I think in this comment you've proposed decent solution https://github.com/rg3dengine/rg3d/issues/155#issuecomment-903024837 . Or maybe I'm missing something?

kazuk:
> I'm curious if such juggling with memory buffers is needed at all. 

This PR implements `no-copy` as possible.

https://github.com/rg3dengine/rg3d/issues/155#issuecomment-903024837 
> I think , too much additional execution cost on this change. please check impact on real world.

If you are not satisfied with the execution cost incurred by a simple implementation, please use this PR.

PR #180, opened 2021-09-14
Deps

martin-t:
Updates the 2 deps that are yellow on deps.rs.

Also removed `extern crate` - it should be no longer necessary in 2018 edition or is there any reason why you used it?


martin-t:
Edit: removed it from everywhere except when needed for `#[macro_use]`. Explanation: https://doc.rust-lang.org/stable/edition-guide/rust-2018/path-changes.html#no-more-extern-crate

Also note i haven't tested any of the changes beyond making sure `cargo build` works, i trust your CI ;)

PR #182, opened 2021-09-14
Fix docs CI

martin-t:
This currently fails (at least we know it works properly), the issues are fixed on another MR.

Also, why use windows-2019? Any particular reason? I think i heard that docs on linux can be a bit faster because windows is slower with lots of files but idk if it applies here or if it matters.


mrDIMAS:
There is no particular reason, initially CI was added by @gideongrinberg and improved by me and @Voodlaz. I was ok with Windows, and just decided to not change it 🤷‍♂️

PR #183, opened 2021-09-14
Cargo audit

martin-t:
Detect vulnerable deps

This should currently fail because there are multiple issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,19 @@ jobs:
              cd examples/wasm
              wasm-pack test --headless --chrome
              wasm-pack test --headless --firefox
+
+  audit:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


martin-t:
In general, some issues are impossible to fix immediately and ok to ignore short term - like warnings that something is unmaintained - there's a flag for ignoring specific advisories - that's better than removing `--deny warnings` because it makes you notice new warns.

Cargo audit and deps.rs works best when your deps use it too :/

Also might be worth putting this in a separate file (e.g. audit.yml) so it doesn't fail all CI and also because we could add another separate badge for this and spread awareness. Which would be awesome.

If i find time tomorrow, i want to try going through some deps and improving their CIs + adding deps.rs and updating their deps but no promises. My "tomorrow" tends to mean "next week" or worse ;)

martin-t:
Argh, i accidentally based this PR on #182 ... let me try again

PR #184, opened 2021-09-14
Cargo audit CI + badge

martin-t:
Detect vulnerable deps

This should currently fail because there are multiple issues.

In general, some issues are impossible to fix immediately and ok to ignore short term - like warnings that something is unmaintained - there's a flag for ignoring specific advisories - that's better than removing --deny warnings because it makes you notice new warns.

Cargo audit and deps.rs works best when your deps use it too :/

Oh and it runs on a timed schedule so even if you don't touch rg3d for over a week and a new vuln pops up, you'll get notified.

This is in a separate file so it doesn't fail *all* CI and also because we can add another separate badge for this and spread awareness. Which would be awesome.

If i find time tomorrow, i want to try going through some deps and improving their CIs + adding deps.rs and updating their deps but no promises. My "tomorrow" tends to mean "next week" or worse ;)

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,28 @@
+name: audit
+
+# This is a separate file so it can have a separate badge in readme
+# and therefore spread awareness of cargo audit a tiny bit.
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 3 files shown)


mrDIMAS:
Sounds great, thanks!

PR #187, opened 2021-09-20
Initial support of `#[derive(Inspect)]`

### LINKED ISSUE #186: Implement `Inspect` proc macro
The new `Inspector` widget will be used to for runtime inspection of an object content in various editors. In rusty-editor it will replace hand-written `Sidebar`, which contains tons of boilerplate code. `Inspector` widget can work only with objects that implements [Inspect](https://github.com/rg3dengine/rg3d/blob/master/rg3d-core/src/inspect.rs#L44) trait, manual implementation of which is very error prone and could be easily offloaded to the proc macro system, in other words - it could be generated automatically for a struct that has `#[derive(Inspect)]` attribute. An example of manual implementation could be found [here](https://github.com/rg3dengine/rg3d/blob/master/src/scene/base.rs#L255). 

There is a list of attributes that should be supported for structure fields (for now it proc macro should be implemented only for structures):

- `#[inspect(skip)]` - do not expose property info
- `#[inspect(name = "<name>")]` - a name override for a field (by default proc macro should take the name of the field, split it at `_` and capitalize all letters)
- `#[inspect(group = "<group>")]` - a group override for a field, by default group should be `Common`

toyboot4e:
Closes #186

# Example use

From [tests/it/inspect.rs](https://github.com/toyboot4e/rg3d/blob/inspect-derive/rg3d-core-derive/tests/it/inspect.rs):

```rust
use rg3d::{
    // maybe add `prelude`?
    core::inspector::{Inspect, ProprertyInfo},
    scene::transform::Transform,
};

#[test]
fn inspect_attributes() {
    #[derive(Default, Inspect)]
    pub struct Data {
        #[inspect(skip)]
        skipped: u32,

        #[inspect(group = "Pos", name = "Super X")]
        x: f32,

        #[inspect(expand)]
        tfm: Transform,

        #[inspect(group = "Pos")]
        y: f32,
    }

    let data = Data::default();

    let expected = vec![
        PropertyInfo {
            name: "Super X",
            group: "Pos",
            value: &data.x,
        },
        PropertyInfo {
            name: "Y",
            group: "Pos",
            value: &data.y,
        },
    ];

    assert_eq!(data.properties()[0..2], expected);
    assert_eq!(data.properties().len(), 2 + data.tfm.properties().len());
}
```

# PR contents

## Preparation
- [x] Restructuring `tests/`  for modulability and to reduce the number of test crates.  
c.f. [Delete Cargo Integration Tests](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html)
- [x] Implement `Debug` and `PartialEq` for `PropertyInfo` (for `assert_eq!` in later tests)

## Requested features in #186
- [x] Support `Struct { named_field: value }`
- [x] `#[inspect(skip)]`
- [x] `#[inspect(name = ..)]`
- [x] `#[inspect(group = ..)]`

## More
- [x] `#[inspect(expand)]`

# TODOs
* Supporting other kinds of structs and enums.
* Improving the `Inspect` trait for performance (if necessary)
* (Just some refactoring to `#[derive(Visit)]`)

### DIFF
--- a/rg3d-core-derive/Cargo.toml
+++ b/rg3d-core-derive/Cargo.toml
@@ -22,4 +22,5 @@ darling = "0.13.0"
 
 [dev-dependencies]
 rg3d-core = { path = "../rg3d-core" }
+rg3d = { path = "../" }
 futures = "0.3.16"
--- a/rg3d-core-derive/README.md
+++ b/rg3d-core-derive/README.md
@@ -1,3 +1,3 @@
 # rg3d-core-derive
 
-Provides the `#[derive(Visitor)]` macro, which implements `rg3d::core::visitor::Visit`.
+Provides the `#[derive(Visit)]` macro, which implements `rg3d::core::visitor::Visit`.
--- a/rg3d-core-derive/src/inspect.rs
+++ b/rg3d-core-derive/src/inspect.rs
@@ -27,7 +27,6 @@ fn impl_inspect_struct(
     );
 
     let impl_body = {
-        // `field.visit(..);` parts
         let props = utils::create_field_properties(
             quote! {  self. },
             field_args.fields.iter(),
--- a/rg3d-core-derive/src/inspect.rs
+++ b/rg3d-core-derive/src/inspect.rs
@@ -26,8 +26,8 @@ fn impl_inspect_struct(
         "#[derive(Inspect) considers only named fields for now"
     );
 
-    let impl_body = {
-        let props = utils::create_field_properties(
+    let prop_vec = {
+        let props = utils::collect_field_props(
             quote! {  self. },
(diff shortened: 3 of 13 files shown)


toyboot4e:
Ah, `#[inspect(expand)]` corresponds to [this line](https://github.com/rg3dengine/rg3d/blob/2429c9d9e859fa5f29578a6cbbea125b929581c8/src/scene/base.rs#L280):

```rust
        properties.extend(self.local_transform.properties().into_iter());
```

Better name to `expand` would exist 🤔

mrDIMAS:
Omg, that's for sure very fast! Great job! I'll check the PR asap!

PR #195, opened 2021-10-03
`#[derive(Inspect)]`: generate property key constants

### LINKED ISSUE #193: `Inspect` proc macro should also generate constants with field names
When using `Inspector` you forced to use string literals to distinguish `PropertyChanged` messages, using raw string literals is very bug prone and this could be improved by using generated string literals which then could be used as `TypeName::FIELD_NAME`- this will result in compiler error if a property name has changed. I propose to generate property names as constants in `impl TypeName`. An example for structures is quite simple:

```rust
#[Inspect]
struct Foo {
  bar: u32
}

// This should be generated
impl Foo {
  pub const BAR: &'static str = "bar";
}
```

Enums are more complex, it should also add variant name prefix to the constant name:

```rust
#[Inspect]
enum Baz {
  Struct { named_field: f32 },
  Tuple(u32, f32),
  Unit,
}

// This should be generated

impl Baz {
  pub const STRUCT_NAMED_FIELD: &'static str = "named_field";
  pub const TUPLE_0: &'static str = "0";
  pub const TUPLE_1: &'static str = "1";
  // Baz::Unit is ignored because it has no fields.
}
```

toyboot4e:
Hi 👋  mr! I'm recently impressed by your realtime attack.. can't estimate your love!

Closes #193. The constants _should_ work as expected, but I didn't test (sorry).

prop-keys

- The constants pullute docs 🤔  But I [didn't find any][D] `#[doc(shrink_by_default)]` attribute.
- Property key constants for private fields are exposed (if it matters, another PR can come).
- Note that there's no key constants for expanded (flatten) field properties (since we can't track the expanded type's fields).

[D]: https://doc.rust-lang.org/rustdoc/the-doc-attribute.html

---

By the way, [this line][L] of `examples/inspector.rs` needs one more field. You might know, but just in case :)

[L]: https://github.com/rg3dengine/rg3d/blob/master/examples/inspector.rs#L141

### DIFF
--- a/rg3d-core-derive/src/inspect.rs
+++ b/rg3d-core-derive/src/inspect.rs
@@ -5,7 +5,7 @@ mod utils;
 
 use darling::{ast, FromDeriveInput};
 use proc_macro2::TokenStream as TokenStream2;
-use quote::{format_ident, quote};
+use quote::{quote};
 use syn::*;
 
 pub fn impl_inspect(ast: DeriveInput) -> TokenStream2 {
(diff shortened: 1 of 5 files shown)


toyboot4e:
Some of `cargo expand --test it` output:

```rust
pub struct STuple(usize);
impl STuple {
    #[allow(missing_docs)]
    pub const F_0: &'static str = "0";
}

#[allow(unused)]
pub enum E {
    Tuple(usize),
    Struct { field: usize },
    Unit,
}
impl E {
    #[allow(missing_docs)]
    pub const TUPLE_F_0: &'static str = "0";
    #[allow(missing_docs)]
    pub const STRUCT_FIELD: &'static str = "field";
}
```

toyboot4e:
Duplicate property key can exist:

```rust
#[derive(Inspect)]
pub enum X {
   V1 { x: usize },
   // variant field with same name
   V2 { x: usize },
}

#[derive(Inspect)]
pub struct A {
    x: usize,
    // expanded field with same name
    #[inspect(expand)]
    xx: X,
}
```

Does it even matter? It's late at night, I'll think about it tomorrow.

toyboot4e:
Just added docstring:

prop-key-docstring

mrDIMAS:
Awesome, that is HUGE help to the project!

PR #206, opened 2021-10-11
update wasm example to use new `set_frame_size()`

w-flo:
I'll be honest, I don't have a wasm build environment set up and I'm low on disk space right now, so this is not even compile-tested. But since I already broke the wasm CI with my last commit, this can't make it worse, right? Let's see...


mrDIMAS:
No problems, thanks for the PR.

PR #242, opened 2021-11-17
Pool u32

martin-t:
- Converted usize to u32 where possible, surprisingly didn't really cause breakage in other parts of the code. Editor and Iapetus are also unaffected.
- `records_get_mut` (and friends) was meant to keep the conversions in one place. Unfortunately rust doesn't have partial borrows yet so it keeps the whole pool borrowed and i couldn't use it in all the places i would have liked - e.g. `free` still needs to do the conversion inside the method itself. It doesn't really matter much but little things like these really annoy the perfectionist in me. Technically `Vec`s in pool could be replaced with some crate that uses u32 instead of usize but it's almost certainly not worth bringing the extra dep.
- Now there's fewer `as` conversions but i still don't feel comfortable that Pool will behave in a sane way if somebody decides to spawn more than 2^32 objects.
- Improved some docs. `put_back` doesn't panic unless there's a bug in `Pool` so removed the Panics heading.

Questions remaining (brought up on discord):
- Change the behavior of `alive_count`? `count` is a proposal for an O(1) solution but behaves differently when reserved objects are in the pool.
- Rename `get_capacity` to `max_len` or something similar?

### DIFF
--- a/rg3d-core/src/pool.rs
+++ b/rg3d-core/src/pool.rs
@@ -1013,6 +1013,11 @@ impl<T> Pool<T> {
         self.iter().count()
     }
 
+    pub fn count(&self) -> u32 {
+        let free = u32::try_from(self.free_stack.len()).expect("free stack length overflowed u32");
+        self.records_len() - free
+    }
+
     #[inline]
     pub fn replace(&mut self, handle: Handle<T>, payload: T) -> Option<T> {
         if let Some(record) = self.records.get_mut(handle.index as usize) {
--- a/rg3d-core/src/pool.rs
+++ b/rg3d-core/src/pool.rs
@@ -364,7 +364,8 @@ impl<T> Pool<T> {
     }
 
     #[inline]
-    pub fn with_capacity(capacity: usize) -> Self {
+    pub fn with_capacity(capacity: u32) -> Self {
+        let capacity = usize::try_from(capacity).expect("capacity overflowed usize");
         Pool {
             records: Vec::with_capacity(capacity),
             free_stack: Vec::new(),
@@ -373,6 +374,16 @@ impl<T> Pool<T> {
 
     fn records_len(&self) -> u32 {
         u32::try_from(self.records.len()).expect("Number of records overflowed u32")
+    }
+
+    fn records_get(&self, index: u32) -> Option<&PoolRecord<T>> {
+        let index = usize::try_from(index).expect("Index overflowed usize");
+        self.records.get(index)
+    }
+
+    fn records_get_mut(&mut self, index: u32) -> Option<&mut PoolRecord<T>> {
+        let index = usize::try_from(index).expect("Index overflowed usize");
+        self.records.get_mut(index)
     }
 
     #[inline]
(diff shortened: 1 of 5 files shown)


martin-t:
re `alive_count`: There's a 3rd option - take_reserve/try_take_reserve and put_back can increment/decrement a counter of reserved objects. Either way, once you decide i will at least document the behavior. I still think it makes sense to count reserved objects as alive since they presumably are alive, just temporarily somewhere else.

mrDIMAS:
Reserved objects are not in the pool, take_reserve moves object out of the pool, but does not put its index to free_stack, so you don't need to take those into account.

mrDIMAS:
I'm not sure if it worth checking if conversion to `u32` has failed (`u32::try_from`), I'm pretty sure that nobody will ever allocate 4 billions objects in a pool. My main concern here is that I want every access operation to be as fast as possible, and adding yet another check could probably harm performance.

martin-t:
> Reserved objects are not in the pool, take_reserve moves object out of the pool, but does not put its index to free_stack, so you don't need to take those into account.

So what does that mean in terms of code? The current `alive_count` doesn't count reserved objects, my new `count` does. So should i:
a) Replace `alive_count`'s implementation by that of my `count`
b) Keep the existing `alive_count` as is and rename `count` to something else? Perhaps `allocated_count` or `total_count`?
c) Expose publicly `records_len` and add `free_stack_len`, keep `alive_count as is and mention these 2 new methods in its docs?

I think b) is a better option than a) because it's not a breaking change and it allows both use cases. Exposing the sized of `records` and `free_stack` might still be interesting from a perf monitoring (and debugging) perspective but i think something like the proposed `total_count` should still be there as a O(1) convenience method.

> I'm not sure if it worth checking if conversion to `u32` has failed (`u32::try_from`), I'm pretty sure that nobody will ever allocate 4 billions objects in a pool. My main concern here is that I want every access operation to be as fast as possible, and adding yet another check could probably harm performance.

I thought you might say that :P

`as` and `usize::try_from(x).expect("test")` compile to
```asm
        mov     eax, edi
        ret
```

`u32::try_from(x).expect("test")` compiles to
```asm
        mov     rax, rdi
        shr     rax, 32
        jne     .LBB2_1
        mov     eax, edi
```
where `.LBB2_1` is the error handling part but it's not on the hot path so it's 3 more instructions (it's similar if the conversion is in a bigger block of code). I think it'll get lost in the noise but i haven't benched it and i don't really mind as long as it's very unlikely to happen in real code.

4 billion objects are not likely to be allocated at the same time, the most likely cause of overflow would IMO be forgetting to put back reserved objects. I did some quick math and it looks like to overflow u32, you'd need to be leaking hundreds of objects per frame for a whole day - i'd be worried more about overflowing the generation first e.g. by spawning and freeing a small number of objects repeatedly in the same arena.

TBH my main worry was about taking a `usize` from outside (in mutiplayer games maybe from an untrusted source) and silently truncating it but since i am converting Pool's whole API to u32, we should be safe.

Soooo if you want, i'll put `as` back for usize to u32 conversions.

mrDIMAS:
Ok, this shouldn't be too much of a problem. The rest looks fine, merging.

PR #245, opened 2021-11-20
Grid optimization

### LINKED ISSUE #243: Optimize Grid's layout
Current implementation of Grid widget layout forcing everything to be re-measured twice, this is very inefficient and in cases of deep nesting of widgets causes dramatic performance issues. There is a [very good implementation with description of algorithms in Avalonia UI ](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Grid.cs#L277) which can be adapted to rg3d-ui's.

This is very important task to make rusty-editor fast on large scenes, currently it is near to unusable when object count on scene is close to few thousands.

mrDIMAS:
Fixes #243

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+
+# Default ignored files
+/workspace.xml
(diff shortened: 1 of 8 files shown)


mrDIMAS:
After a few days of hard work I managed to improve performance 1000x times. I measured performance using Visual Studio profiler in Release build of rusty-editor on large scene with few thousands objects, in total it had ~23000 widgets. Before optimization grid layout took around 60% of time, now it takes 0.06% which gives 1000x times performance improvement.

PR #246, opened 2021-11-23
Restyle inspector

mrDIMAS:


mrDIMAS:
So this is the new look of the inspector, it is much nicer compared to old one:
new

Old one for reference:
old

PR #252, opened 2021-12-03
Replace DefaultHasher with FxHasher

martin-t:
I didn't do a thorough check but it looks like none of these hashes survive after engine exit so changing them should be ok. Besides, DefaultHasher doesn't guarantee stable hashes either.

### DIFF
--- a/rg3d-core/src/profiler.rs
+++ b/rg3d-core/src/profiler.rs
@@ -4,9 +4,8 @@
 
 #![allow(dead_code)]
 
-use fxhash::{FxHashMap, FxHashSet};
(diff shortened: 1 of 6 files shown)


martin-t:
I did some quick measurements of this MR and the previous one using 3rd_person and TBH i don't see much difference.

FPS using quality 1,2,3,4:
Originally: 26,27,95,oscillating around 290
Now: 26,27,95-100,oscillating around 290
So maybe a few more FPS at most. Might be interesting if the engine kept more long term stats or remembered extremes (especially drops in FPS). There might also be scenes which use hashmaps more, 3rd_person is after all pretty simple.

PR #257, opened 2021-12-14
Allow disabling frustum_culling

Nibor62:
Hi,

Not sure disabling frustum culling will be the right solution in my case, but here is a pr in case you are interested.

Still a bit wip
Not tested much in depth
Haven't added any kind of automated tests (is there any anyway ?)
Also, where should doc for this field by added ?

### DIFF
--- a/src/scene/base.rs
+++ b/src/scene/base.rs
@@ -595,7 +595,7 @@ impl Base {
         self.frustum_culling
     }
 
-    /// Sets the frustum_culling flag
+    /// Sets whether to use frustum culling or not
     pub fn set_frustum_culling(&mut self, frustum_culling: bool) {
         self.frustum_culling = frustum_culling;
     }
@@ -628,6 +628,7 @@ impl Visit for Base {
         self.tag.visit("Tag", visitor)?;
         let _ = self.properties.visit("Properties", visitor);
         let _ = self.physics_binding.visit("PhysicsBinding", visitor);
+        self.frustum_culling.visit("FrustumCulling", visitor)?;
 
         visitor.leave_region()
     }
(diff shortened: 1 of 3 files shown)


src/scene/base.rs
```diff
@@ -342,6 +342,7 @@ pub struct Base {
     /// A set of custom properties that can hold almost any data. It can be used to set additional
     /// properties to scene nodes.
     pub properties: Vec<Property>,
+    frustum_culling: bool,
```
mrDIMAS, line 345: You should also serialize this field. Go to `impl Visit for Base` and add somewhere at the end `let _ = self.frustum_culling.visit("FrustumCulling", visitor);`

mrDIMAS (changes requested):

src/scene/base.rs
Nibor62: Does the visitor thing also handle de-serialization ?

Nibor62:
Do you need more doc and / or tests (beyond some eye check ?)

src/scene/base.rs
```diff
@@ -639,6 +651,7 @@ impl Visit for Base {
         self.tag.visit("Tag", visitor)?;
         self.physics_binding.visit("PhysicsBinding", visitor)?;
         let _ = self.properties.visit("Properties", visitor);
+        self.frustum_culling.visit("FrustumCulling", visitor)?;
```
mrDIMAS, line 654: `let _ = self.frustum_culling.visit("FrustumCulling", visitor);`, do not use `?` operator because it will break backward compatibility with previous versions.

mrDIMAS (changes requested):

src/scene/base.rs
mrDIMAS: Visitor is a serializer+deserializer so yes :)

PR #278, opened 2022-01-18
Add `#[inspect(getter = "<path>")]` attribute #277

### LINKED ISSUE #277: Add a `getter` attribute for `Inspect` proc macro
Sometimes there is a need for custom getter for a property. For example - https://github.com/FyroxEngine/Fyrox/blob/master/src/scene/transform.rs#L99.  Currently there is no way to specify a getter function for `Inspect` proc macro. I think something like this should work:

```rust
#[derive(Inspect)]
struct Foo {
  #[inspect(getter = get_bar)]
  bar: u32
}

impl Foo {
  fn get_bar(&self) -> &u32 {
    &self.bar
  }
}
```

This code should generate something like this:

```rust
...
impl Inspect for Foo {
    fn properties(&self) -> Vec<PropertyInfo<'_>> {
        vec![
          PropertyInfo {
              owner_type_id: TypeId::of::<Self>(),
              name: "bar",
              display_name: "Bar",          
              value: self.get_bar(), // Or &self.bar if getter is not specified.
              read_only: false,
              min_value: None,
              max_value: None,
              step: None,
              precision: None,
              description: "".to_string(),
          },
       ]
    }
  }
}
```

toyboot4e:
Resolves #277.

Example:

```rust
use std::ops::Deref;

#[derive(Inspect)]
struct Foo {
  #[inspect(getter = get_ref)]
  bar: u32

  #[inspect(getter = Deref::deref)]
  baz: TemplateValue<Vec3>,
}

fn get_ref(bar: &u32) -> &u32 { bar }
```

I was confusing in #277, but the getter function takes the reference of the field, not the inspctable type (`&self`).

> We could hook custom user function, but this is not the case for it :)


mrDIMAS:
That was quick, thanks a lot!

Issue #285, opened 2022-01-26
Add `is_modified` field for `PropertyInfo` + proc macro support


mrDIMAS:
Editor support for https://github.com/FyroxEngine/Fyrox/issues/282 requires an ability to get an `is_modified` flag at runtime from a PropertyInfo of a particular property to show it in Inspector as a small marker. The `is_modified` flag will also be used to revert value to the parent's value. The value of the flag should be taken by a user-defined function. For example:

```rust
use crate::core::inspect::{Inspect, PropertyInfo};
use crate::scene::variable::TemplateVariable;
use std::any::TypeId;

#[derive(Inspect)]
struct Foo {
    #[inspect(is_modified = "is_modified")]
    value: TemplateVariable<f32>,
}

// Generated code should look like this.
impl Inspect for Foo {
    fn properties(&self) -> Vec<PropertyInfo<'_>> {
        vec![PropertyInfo {
            owner_type_id: TypeId::of::<Self>(),
            name: "value",
            display_name: "value",
            value: &self.value,
            read_only: false,
            min_value: None,
            max_value: None,
            step: None,
            precision: None,
            description: "".to_string(),
            // If is_modified proc macro argument is not defined, then is_modified should be false
            is_modified: self.value.is_modified(),
        }]
    }
}
```

mrDIMAS:
@toyboot4e is this easy to implement? Also any thoughts and improvements to the design are very welcome :)

toyboot4e:
It's easy 👍 I've sent a PR

PR #286, opened 2022-01-26
add `PropertyInfo::is_modified` #285

toyboot4e:
Closes [#285](https://github.com/FyroxEngine/Fyrox/issues/285):

```rust
#[derive(Inspect)]
struct S(#[inspect(is_modified = "Clone::clone")] bool);
```

I personally preferred to allow any path, but changing it to method name is also easy :)

### DIFF
--- a/fyrox-core-derive/src/inspect/args.rs
+++ b/fyrox-core-derive/src/inspect/args.rs
@@ -97,6 +97,12 @@ pub struct FieldArgs {
     /// Description of the property.
     #[darling(default)]
     pub description: Option<String>,
+
+    /// `#[inspect(getter = "<method_name>")]`
+    ///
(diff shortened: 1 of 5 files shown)


toyboot4e:
Oh `is_modified = FieldType::method` is verbose 🤔

toyboot4e:
So I changed the syntax to the original one that takes method name:

```rust
#[derive(Inspect)]
struct S(#[inspect(is_modified = "clone")] bool);
```

Today I'm feeling like `getter` should use method name (e.g. `deref`), too 🤔

mrDIMAS:
Both options are fine, but I prefer shorter option 😉 . Thanks for the quick implementation, your help is immeasurable, great job!

PR #291, opened 2022-02-06
Make everything called from Engine::update public

martin-t:
As discussed on discord, this is needed to be able to split physics and UI updates.

I also renamed renderer's `update` to `update_caches` to make it clearer what it does (i thought it did more and found it surprising renderer updated before physics).

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -279,7 +279,7 @@ impl Engine {
         let window_size = Vector2::new(inner_size.width as f32, inner_size.height as f32);
 
         self.resource_manager.state().update(dt);
-        self.renderer.update(dt);
+        self.renderer.update_caches(dt);
         self.handle_model_events();
 
         for scene in self.scenes.iter_mut().filter(|s| s.enabled) {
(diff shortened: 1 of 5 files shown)


src/engine/resource_manager/mod.rs
```diff
@@ -370,7 +370,8 @@ impl ResourceManagerState {
         containers.curves.destroy_unused();
     }
 
-    pub(in crate) fn update(&mut self, dt: f32) {
+    /// Reload resources if they have changed in disk.
```
mrDIMAS, line 373: I think this comment is misleading, what `update` really does is it updates resource containers *and* performs hot reloading. Update of resource container removes resources that weren't used for some time (afaik 60 seconds by default).

Also it is worth noting that this method should not be called manually if one use `Engine::update`

src/engine/mod.rs
```diff
@@ -299,7 +299,8 @@ impl Engine {
         self.ui_time = instant::Instant::now() - time;
     }
 
-    pub(crate) fn handle_model_events(&mut self) {
+    /// Handle hot-reloading of resources.
```
mrDIMAS, line 302: It is worth noting that this method should not be called manually if one use `Engine::update`

mrDIMAS (changes requested):

martin-t:
Ok, i updated the comment for all 3 newly public methods.

PR #292, opened 2022-02-06
Add option to define custom resource loaders

gzp-crey:
The loaders for the resources can be customized when the engine is created:
- Engine requires the ResourceManager to be created explicitly.
- The default loader and ResourceLoader traits were made public
- Some minor API changes were made on the ResourceLoader to easy its usage
  - removed the associated Future type
  -  `&self` instead of `&mut self`
  - Send for not wasm targets
  
A new example was added. Later it could be improved to add real loader (ex load model from other file formats)

It seems as this is only a first step because the resource data (ModelData, TextureData etc.) has no functionality yet to create them "by hand".

A next step could be to move the fbx loader into it's own crate and it'd prove that complex scene can be loaded without the need of any internals of the engine. But it requires some refactor - I don't see yet -  to avoid a circular dependency (engine depends on fbx to enable the loader, and fbx depends on engine for the scene features).

### DIFF
--- a/examples/save_load.rs
+++ b/examples/save_load.rs
@@ -17,7 +17,17 @@
 
 pub mod shared;
 
-use crate::shared::{create_ui, fix_shadows_distance, Game, GameScene, LocomotionMachine, Player};
+use crate::shared::{create_ui, fix_shadows_distance, create_scene_async, Game, GameScene, LocomotionMachine, Player};
+use fyrox::engine::Engine;
+use fyrox::engine::resource_manager::ResourceManager;
+use fyrox::engine::resource_manager::container::event::ResourceEventBroadcaster;
+use fyrox::engine::resource_manager::loader::model::ModelLoader;
+use fyrox::engine::resource_manager::loader::texture::TextureLoader;
+use fyrox::engine::resource_manager::loader::{ResourceLoader, BoxedLoaderFuture};
+use fyrox::engine::resource_manager::options::ImportOptions;
+use fyrox::event_loop::EventLoop;
+use fyrox::resource::model::{ModelImportOptions, Model};
+use fyrox::resource::texture::{TextureImportOptions, Texture};
 use fyrox::scene::SceneLoader;
 use fyrox::{
     core::{
@@ -37,7 +47,9 @@ use fyrox::{
     },
 };
 use fyrox_core::futures::executor::block_on;
+use std::marker::PhantomData;
 use std::path::Path;
+use std::sync::Arc;
 
 // Start implementing Visit trait for simple types which are used by more complex.
 // At first implement trait for LocomotionMachine.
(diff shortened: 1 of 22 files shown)


examples/save_load.rs
```diff
@@ -17,7 +17,19 @@
 
 pub mod shared;
 
-use crate::shared::{create_ui, fix_shadows_distance, Game, GameScene, LocomotionMachine, Player};
+use crate::shared::{
```
mrDIMAS, line 20: Please make a separate example for custom resource loaders. `save_load` example is a about serialization, not about resources.

src/engine/mod.rs
```diff
@@ -139,17 +141,20 @@ impl Engine {
...
     /// let window_builder = WindowBuilder::new()
     ///     .with_title("Test")
     ///     .with_fullscreen(None);
-    /// let mut engine: Engine = Engine::new(window_builder, &evt, true).unwrap();
+    /// let mut engine: Engine = Engine::new(window_builder, resource_manager_builder, &evt, true).unwrap();
     /// ```
     #[inline]
     pub fn new(
         window_builder: WindowBuilder,
+        resource_manager_builder: ResourceManagerBuilder,
```
mrDIMAS, line 157: This should accept `ResourceManager` itself, not a builder. First parameter (`window_builder`) is passed as builder because it is used by glutin's API.

src/engine/resource_manager/loader/curve.rs
```diff
@@ -10,15 +11,14 @@ use crate::{
...
 
 impl ResourceLoader<CurveResource, CurveImportOptions> for CurveLoader {
-    type Output = BoxedLoaderFuture;
-
     fn load(
-        &mut self,
+        &self,
         curve: CurveResource,
         _default_import_options: CurveImportOptions,
+        _resource_manager: ResourceManager,
```
mrDIMAS, line 18: Do not pass resource manager to `load`, this was removed in one of latest commits. Loaders should acquire their own copy of ResourceManager explicitly, as it was already done for ModelLoader.

src/engine/resource_manager/loader/curve.rs
```diff
@@ -10,15 +11,14 @@ use crate::{
 pub struct CurveLoader;
 
 impl ResourceLoader<CurveResource, CurveImportOptions> for CurveLoader {
-    type Output = BoxedLoaderFuture;
-
     fn load(
-        &mut self,
+        &self,
```
mrDIMAS, line 15: Please leave this as `&mut self`, loader potentially can mutate its state during the load.

src/engine/resource_manager/mod.rs
```diff
@@ -30,28 +31,92 @@ use notify::DebouncedEvent;
...
+
+pub struct ResourceManagerBuilder {
+    texture_loader: Box<dyn ResourceLoader<Texture, TextureImportOptions>>,
+    model_loader: Box<dyn ResourceLoader<Model, ModelImportOptions>>,
+    sound_buffer_loader: Box<dyn ResourceLoader<SoundBufferResource, SoundBufferImportOptions>>,
+    shader_loader: Box<dyn ResourceLoader<Shader, ShaderImportOptions>>,
+    curve_loader: Box<dyn ResourceLoader<CurveResource, CurveImportOptions>>,
+}
+
+impl ResourceManagerBuilder {
```
mrDIMAS, line 65: Builder should have `build` method that returns `ResourceManager`.

mrDIMAS (changes requested):
Also, could you please document your changes? CI failing because of this.

src/engine/resource_manager/loader/curve.rs
gzp-crey: > Do not pass resource manager to `load`, this was removed in one of latest commits. Loaders should acquire their own copy of ResourceManager explicitly, as it was already done for ModelLoader.

That would be an issue during ResourceManager creation: In the current API resource manager is created later as the loaders. In the non-configurable current solution in the new The resource manager is crated than the loaders and finally they are "glued" together (https://github.com/FyroxEngine/Fyrox/blob/c24bfa5679161625c7edad07d7df93cb75979522/src/engine/resource_manager/mod.rs#L97). 
So either the whole state setup have to be moved outside of ResourceManager::new  or ResourceManager may require functions to alter the loaders once it's been created. The latter may align better to provide a ResourceManager instead of a builder for the Engine.

mrDIMAS: Yes, a set of functions to replace loaders is a good alternative.

src/engine/resource_manager/loader/curve.rs
mrDIMAS: You absolutely right, I just didn't thought about this, sorry.

gzp-crey: I'd have not thought of it prior trying to implement the sample wrapping loaders :).

src/engine/mod.rs
gzp-crey: done

src/engine/resource_manager/mod.rs
gzp-crey: There was no need for the builder after the change as engine requires a ResourceManager (and not a builder)

mrDIMAS (approved):
Alright, seems good to me.

PR #301, opened 2022-02-17
Add FontBuilder

therealbnut:
I wanted to make the default font size a little bigger because I had trouble reading it on my screen. I searched the discord and there didn't seem to be a good way to do it yet.

I decided the best way would be to make a `FontBuilder` that also encapsulates the idea of the `DEFAULT_FONT`.

This isn't a general solution to the issue of dynamic font size changes, and font atlases etc. but it might lower the barrier to getting started.

I also changed `Font:: from_memory` to take `Deref<Target = [u8]>` so you don't need to copy it into a `Vec` if you already have the slice. I don't think this will break existing code, but it may give a clippy warning (if they use it) telling them they don't need to call `.into()`.

### DIFF
--- a/fyrox-ui/src/ttf.rs
+++ b/fyrox-ui/src/ttf.rs
@@ -4,6 +4,7 @@ use crate::{
 };
 use fxhash::FxHashMap;
 use std::{
+    borrow::Cow,
     fmt::{Debug, Formatter},
     ops::{Deref, Range},
     path::Path,
@@ -139,7 +140,7 @@ impl Font {
     }
 
     pub fn from_memory(
-        data: Vec<u8>,
+        data: impl Deref<Target = [u8]>,
         height: f32,
         char_set: &[Range<u32>],
     ) -> Result<Self, &'static str> {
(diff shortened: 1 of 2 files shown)


fyrox-ui/src/ttf.rs
```diff
@@ -294,3 +295,62 @@ impl Font {
...
+
+    /// Sets the desired character set for the produced font.
+    #[inline]
+    pub fn with_char_set(mut self, char_set: impl Into<Cow<'a, [Range<u32>]>>) -> Self {
+        self.char_set = Some(char_set.into());
+        self
+    }
+
+    /// Creates a new font from the data at the specified path.
+    pub async fn build_from_file(self, path: impl AsRef<Path>) -> Result<Font, &'static str> {
```
therealbnut, line 330: I decided to make multiple build methods because of the mix of `async` and sync building.

You could also do something like this:
```rust
pub enum FontSource<'a> {
    BuiltIn,
    Path(&'a Path),
    Memory(&'a [u8]),
}

pub fn with_source(self, source: FontSource) -> Self { ... }

pub fn build(self) -> ... {}
```

Or something like this ⬇️ - but I think it's better if the builder is lazy:
```rust
pub async fn with_data_from_file(self, path: impl AsRef<Path>) -> Result<Self, &'static str> { ... }
```

A more complicated way to do it, but with perhaps a more familiar API is:
```rust
pub struct BuiltInFontSource;
pub struct FontBuilder<'a, Source = BuiltInFontSource> { source: Source, ... }
impl<'a> FontBuilder<'a, BuiltInFontSource> {
    pub fn with_data_from_memory(self, data: impl Deref<Target = [u8]>) -> FontBuilder<'a, &'a [u8]> { ... }
    pub fn with_data_from_file(self, path: impl AsRef<Path>) -> FontBuilder<'a, &'a Path> { ... }
    pub fn build(self) -> Result<Font, &'static str> { ... }
}
impl<'a> FontBuilder<'a, &'a [u8]> {
    pub fn build(self) -> Result<Font, &'static str> { ... }
}
impl<'a> FontBuilder<'a, &'a Path> {
    pub async fn build(self) -> Result<Font, &'static str> { ... }
}
```

mrDIMAS: Current solution is perfectly fine, thanks!

mrDIMAS (approved):

fyrox-ui/src/ttf.rs
therealbnut: Thanks for the review!

PR #315, opened 2022-04-29
Internal: change property key name to consider enum variant's name

toyboot4e:
This PR changes the property key constants for enums generated with `#[derive(Inspect)]`.

# Example

Let's consider this code:

```rust
#[derive(Inspect)]
enum E {
    A { field: usize },
    B { field: usize },
}
```

Before this PR, it generates:
```rust
impl E {
    pub const A: &'static str = "field";
    pub const B: &'static str = "field";
}
```

After this PR, it generates:
```rust
impl E {
    pub const A: &'static str = "A.field";
    pub const B: &'static str = "B.field";
}
```

# Future use

With runtime reflection, we might refer to enum fields like `joint.BallJoint.local_anchor1`.

### DIFF
--- a/fyrox-core-derive/src/inspect/utils/prop_keys.rs
+++ b/fyrox-core-derive/src/inspect/utils/prop_keys.rs
@@ -6,12 +6,36 @@ use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use syn::*;
 
-use crate::inspect::{args, utils};
+use crate::inspect::args;
 
-/// Returns a list of `pub const [VARIANT_]FIELD: &'static str = "key_value"`;
+/// `pub const [VARIANT_]FIELD: &'static str = "key";`
+pub fn prop_keys_impl(ty_args: &args::TypeArgs) -> TokenStream2 {
+    let ty_ident = &ty_args.ident;
+    let (impl_generics, ty_generics, where_clause) = ty_args.generics.split_for_impl();
+
+    let prop_keys = self::quote_prop_keys(ty_args);
+
+    quote! {
+        /// Property key constants
+        impl #impl_generics #ty_ident #ty_generics #where_clause {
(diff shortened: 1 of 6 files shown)


fyrox-core-derive/src/inspect/utils/prop_keys.rs
```diff
@@ -6,12 +6,36 @@ use proc_macro2::TokenStream as TokenStream2;
...
+
+/// List of `pub const <ident> = <name>;`
+///
+/// # Definition format
+///
+/// | Type         | Identifier                | Name                     |
+/// |--------------|---------------------------|--------------------------|
+/// | Struct       | `FIELD_NAME`              | `field_name`             |
+/// | Unit struct  | `F_<number>`              | `<number>`               |
+/// | Tuple struct | `VARIANT_NAME_FIELD_NAME  | `VariantName.field_name` |
```
mrDIMAS, line 34: Looks like `Tuple struct` should be swapped with `Enum`, because otherwise it looks odd to have `VARIANT_NAME_F_<number>` for enumerations 🤔

mrDIMAS (approved):
Everything looks good, there is one minor issue with a comment, but I could be wrong about it.

fyrox-core-derive/src/inspect/utils/prop_keys.rs
toyboot4e: Right! I've updated it to this:

```rust
/// | Struct         | `FIELD_NAME`              | `field_name`             |
/// | Struct (tuple) | `F_<number>`              | `<number>`               |
/// | Enum (struct)  | `VARIANT_NAME_FIELD_NAME  | `VariantName.field_name` |
/// | Enum (tuple)   | `VARIANT_NAME_F_<number>` | `VariantName.<number>`   |
```

Issue #330, opened 2022-07-18
Add reflection


mrDIMAS:
There is a need in a runtime reflection to reduce amount of boilerplate code needed to handle changed properties in the editor. Reflection should support setting a value using its path (like "foo.bar.baz[0].foobar").

toyboot4e:
mrDIMAS on Discord pointed some boilerplate codes:

- Define command for each field:
https://github.com/FyroxEngine/Fyrox/blob/7b2b76f4928b308e8dd2da1efef6e4544da702cf/editor/src/scene/commands/camera.rs#L19

- Handle each property change:
https://github.com/FyroxEngine/Fyrox/blob/7b2b76f4928b308e8dd2da1efef6e4544da702cf/editor/src/inspector/handlers/node/base.rs#L19

PR #333, opened 2022-07-24
Add `Reflect` trait

toyboot4e:
Related to #330

# Usage

- How can I use `Reflect` traits?
See [reflect.rs](https://github.com/FyroxEngine/Fyrox/pull/333/files#diff-8413c04cc9ffce40e9883d061d4e99fc378357e401c9768b667c6fe427bc0c31) and the [test cases](https://github.com/FyroxEngine/Fyrox/pull/333/files#diff-b58a80f4e9ef6e55b47684800e35480b30419445acb57ef74a7846c8460bebc7).
- How can I implement `Reflect` for more types?
See [external_impls.rs](https://github.com/FyroxEngine/Fyrox/pull/333/files#diff-9aa94bd89b5abedaee324273cc4eddd3ec8627c175dee8aeffe6dffacb5fb3b5) as an example. It uses the derive macro to implement `Reflect` trait for external types.

# PR contents

This PR does the followings:

- Add `Reflect` trait
- Move property key constant generation from `#[derive(Inspect)]` to `#[derive(Reflect)]`
- Add a bunch of `#[derive(Reflect)]` s

Note that the compatibilities of `Inspect` property path and  `Reflect` path are NOTE tested.

## Update

More changes:

- Add `Reflect::as_list*`
Downsides: now `Reflect` is implemented for only `Vec<T>` where `T: Reflect`
- Handle reflect path with `usize` indices
e.g. `assert_eq!(a.get_resolve_path("xs[0].data[1]"), Ok(&1usize))`)
- Add `BaseNodeTrait: Reflect` boundary and remove `BaseNodeTrait::as_any*`
This change may be out of the PR's scope, sorry!
- Remove `#[inspect(name = ..)]`
Since property path and reflect path must match.

Note:

- Why not just copy the awesome `bevy_reflect`?
To make it just work! I'm happy to see my macros replaced with better ones though.
- Can it replace scene commands?
Probably, but `Reflect` API may need to be tweaked to support swap commands.


mrDIMAS:
Incredible work and very well done, thank you! I'm sure it will cut 90%+ of boilerplate code in the editor, however I need to test it how it will work with commands.

PR #359, opened 2022-09-05
Implement property inheritance via reflection

mrDIMAS:
This PR implements automatic property inheritance using reflection. Property inheritance is a mechanism that allows to inherit values of properties from prefabs (including nesting prefabs). It is very powerful and allows to create prefab instances and slightly modify them (changing position, some other properties, etc.) while having rest of the properties to be inherited, which allows you to change prefabs and these change will reflect in instances.

Key features of the PR:

- Ability to revert modified properties to their parent values in prefab (small `<` button near modified property in Inspector)
- Automatic hierarchical property inheritance - it means that you just add `InheritableVariable<T>` field in your script/node and it will participate in inheritance automatically.
- No need to add additional code gen parameters for most of the cases
- `#[inspect(deref)]` and `#[reflect(deref)]` are not needed anymore for `InheritableVariable<T>` fields
- Removed `Inherit` proc macro (sorry @toyboot4e 🙏 )
- Remove `DirectlyInheritableEntity` trait and all boilerplate code for property inheritance
- Added `InheritablePropertyEditor` for `InheritableVariable<T>` fields which acts like a wrapper for inner type, but adds some control elements to property.


toyboot4e:
Cool!

PR #369, opened 2022-09-27
Remove `?` operators from proc-macro generated code

toyboot4e:
This is an internal clean up PR. Hopefully it can also improve the compile time a bit.

From Twitter:

- [MiniJinja author said](https://twitter.com/mitsuhiko/status/1574111459361132549) avoiding `?` improved both the compile time and the runtime performances.
- [dtolnay said](https://twitter.com/davidtolnay/status/1574266317762224128) they don't use `?` operators in `serde` and `serde-json` "because it is too slow".
- [serde has their own `try!` macro](https://github.com/serde-rs/serde/blob/master/serde_derive/src/try.rs) that avoids type conversion
- [Relevant issue on the rust repo](https://github.com/rust-lang/rust/issues/37939)

### DIFF
--- a/fyrox-core-derive/src/reflect.rs
+++ b/fyrox-core-derive/src/reflect.rs
@@ -300,7 +300,11 @@ fn gen_impl(
             }
 
             fn set(&mut self, value: Box<dyn Reflect>) -> Result<Box<dyn Reflect>, Box<dyn Reflect>> {
-                let this = std::mem::replace(self, value.take()?);
+                let value = match value.take() {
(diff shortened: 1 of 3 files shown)


lenscas:
> CI build time on Windows changed from 8m 45s to 8m 5s. I think it doesn't mean anything though -- maybe they were run on different machine?

I don't think you can use github actions to measure performance as it will probably have quite a bit of variance as it is running on a VM at best and there are probably quite a few other VM's running on the same actual hardware doing their own things

PR #414, opened 2022-12-20
add a bit more information to panic on writing file

lenscas:
hastly made through githubs interface. Having checked if it compiled as a result. Should at least give you a start though :)


lenscas:
> error: could not compile `fyrox` due to previous error
>
> Caused by:
>  There is not enough space on the disk. (os error 112)

Not much I can do about that...

Issue #222, opened 2021-11-05
Headless (dedicated) mode for multiplayer game servers


martin-t:
I am not completely sure what i really want here, i am still figuring things out but i'll create an issue now to share my findings, maybe other people have also tried writing multiplayer game servers and will chip in. I'll keep updating the issue as i have a clearer idea of what i want from rg3d and maybe we can come up with a decent API for it.

- There should be a way to use rg3d without creating a window. `Engine::new` expects a window builder and an event loop. The window builder is completely unnecessary and i *think* i'd prefer to write my own event loop as well since all it's gonna do is respond to network anyway. (Long term maybe also stdin/out for quick server config changes and filesysten changes for hot reloading.)
- Most of `Engine`'s fields are not necessary - only `resource_manager`, `scenes` and `scenes2d` are needed on a server.
- It would be nice to load only the scene graph and geometry from assets for collision detection. Textures are not necessary, not loading them could save memory.

---

- Not necessarily related but mentioning it here in case it affects API design: Multiplayer game clients should send (some kinds of) input to server immediately, not once per frame. That means blocking the main thread with winit's event loop as little as possible and doing all work on a separate thread. This is *probably* already achievable without changes to rg3d (by spawning a separate thread, passing input to it using channels, doing engine updates and rendering from that thread).

martin-t:
I wanted a way to run my fyrox game on github's CI. This has 2 issues:
- There are no graphics
  - This can be solved on linux by using xvfb and installing some opengl libs (libegl1-mesa libegl1) and running inside xvfb: https://github.com/rustcycles/rustcycles/blob/3d701480c7574a5f645344282dd25babf65c79fb/.github/workflows/ci.yml#LL56C22-L56C22. 
  - Idk how to solve it on windows and mac.
- There is no audio. I couldn't find a way to get around that, the game would simply crash so i made a PR that offers an option to use a dummy audio device.

mrDIMAS:
This is now possible to do by not calling `Engine::initialize_graphics_context` when using the engine in a "framework" mode (read - with custom game loop).

PR #416, opened 2022-12-23
Minimal support for headless mode

martin-t:
Related to #222 - see latest comment.

I am not sure if the field should be named headless because it just disables sound, not graphics but i figured if adding this field is gonna be a breaking change, might as well call it that so that when somebody figures out a way to disable window creation, it's not another breaking change later.

Basically i thought i'd need to implement a full headless mode but for my usecase (running on CI) it turns disabling sound is all that's needed. I currently don't have any intention of looking into how to implemented a full headless mode because i suspect it would be a lot of effort. Nonetheless, i think this PR is very useful to be able to run fyrox games on CI, the only issue for me is whether i should keep calling that field `headless` or change the name to `noaudio`, `no_sound` or something similar.

### DIFF
--- a/editor/src/lib.rs
+++ b/editor/src/lib.rs
@@ -511,6 +511,7 @@ impl Editor {
             serialization_context,
             events_loop: event_loop,
             vsync: true,
+            headless: false,
         })
         .unwrap();
 
--- a/examples/custom_loader.rs
+++ b/examples/custom_loader.rs
@@ -203,6 +203,7 @@ fn main() {
         serialization_context,
         events_loop: &event_loop,
         vsync: false,
+        headless: false,
     })
     .unwrap();
 
--- a/examples/custom_loop.rs
+++ b/examples/custom_loop.rs
@@ -31,6 +31,7 @@ fn main() {
         serialization_context,
         events_loop: &event_loop,
         vsync: true,
+        headless: false,
     })
     .unwrap();
 
--- a/examples/inspector.rs
+++ b/examples/inspector.rs
@@ -118,6 +118,7 @@ fn main() {
         serialization_context,
         events_loop: &event_loop,
         vsync: false,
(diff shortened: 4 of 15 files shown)


martin-t:
Oh and i added `SoundEngine::new_headless()` because i assume it's better than `without_device`. It looks like even the dummy device does mixing so i assume it has some side effects (plus the spawned thread). `without_device` says the user must periodically call `render` so i assume calling `render` is important even if we don't play any sound - maybe it consumes some buffer or something? I didn't look too deeply into it though, it's possible that `without_device` is sufficient for this usecase, let me know.

mrDIMAS:
I think this approach with flags is quite hacky. Instead it should be done via conditional compilation, that will completely exclude renderer and sound backends from the engine. Actually, I want to experiment with this, because I'm so tired of animations and stuff. This should be fun to do.

martin-t:
I don't think conditional compilation is a good option. That would mean having to compile the engine twice for client and server (increasing CI time) and always producing 2 separate final binaries (increasing incremental build time which is essential to keep down for gamedev). My game only has 1 binary that allows running as client or server or both in one process specifically because of this.

mrDIMAS:
Good point. I poked conditional compilation a bit today and I see that it will add quite a lot of compilation conditions for the game code, making it quite annoying to write and maintain. I guess then the overall idea of the PR is fine, I just need to look at code again.

As for further changes, I think wrapping the renderer into `Option<Renderer>` would be good and simple solution to disable it too for the server mode. The `winit`'s event loop can be used for the server logic without any problems, it is possible to run it without a window. So making `EngineInitParams::window_builder` and related fields (such as window and stuff in the `Engine`) optional is a way to go. I think it is not so hard to do, I'll take a look at it tomorrow.

mrDIMAS (approved):

PR #448, opened 2023-04-15
Method to link nodes while keeping child's global position and rotation.

IceGuye:
This method is helpful on senerio like putting a nail into a wood, or arrow hitting a target and stuck into the target.

### DIFF
--- a/src/scene/graph/mod.rs
+++ b/src/scene/graph/mod.rs
@@ -450,6 +450,25 @@ impl Graph {
         self.pool[parent].children.push(child);
     }
 
+    /// Links specified child with specified parent while keeping the
+    /// child's global position and rotation.
+    #[inline]
+    pub fn link_nodes_keep_global_position_rotation (
+        &mut self,
+        child: Handle<Node>,
+        parent: Handle<Node>) {
+        let parent_transform_inv = self.pool[parent]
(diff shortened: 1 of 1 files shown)


src/scene/graph/mod.rs
```diff
@@ -450,6 +450,25 @@ impl Graph {
...
+        child: Handle<Node>,
+        parent: Handle<Node>) {
+        let parent_transform_inv = self.pool[parent]
+            .global_transform().try_inverse().unwrap_or_default();
+        let child_transform = self.pool[child].global_transform();
+        let relative_transform = parent_transform_inv * child_transform;
+        let local_position = relative_transform.position();
+        let local_rotation = UnitQuaternion::from_matrix(
+            &relative_transform.basis());
+        self.pool[child].local_transform_mut().set_position(local_position);
```
mrDIMAS, line 467: You borrowing `child` two times here, you should do `self.pool[child].local_transform_mut().set_position(local_position).set_rotation(local_rotation)` instead.

mrDIMAS (changes requested):

src/scene/graph/mod.rs
IceGuye: This has been fixed.

PR #451, opened 2023-04-26
log location changed; fixing editor

idrisz19:
Change made due to [logger changing location](https://github.com/FyroxEngine/Fyrox/commit/0a05c4eecd9952c208905bb6d70f936350a43a02). The editor fails without this change.


mrDIMAS:
Good find, I completely forgot about `fyrox-template`. However, the generated code should be changed too, otherwise a generated project won't compile. I'll do it myself.

Issue #360, opened 2022-09-05
Editor high CPU usage


hw762:
It seems that the Fyrox editor eats 100% cpu (one core) even at idle. Is something spinning a loop? It would be nice if this usage can be reduced.

martin-t:
Make sure you've compiled the editor in release mode.

Apart from that I am also noticing 100% CPU usage in my game, seems to only happen after it's been left minimized for a while.

vpochapuis:
I have experienced a similar CPU Usage issue.
Issue to reproduce:

- Start the Editor in Release mode
- Click `Play` in the editor
- Do not quit the game using the game window, but the Editor's "Stop" button

Result:
CPU percentage usage increases.

I can repeat this indefinitely until my total CPU load is 100%.

One of the weird aspect of this issue, the CPU usage increase doesn't happen when the game is started, it happens when the game is closed.

Here is the CPU usage during the following procedure:
- Start and stop the game from the editor several times
- Close the editor

![image](https://user-images.githubusercontent.com/75721408/208374158-489a1a83-dfe6-4d21-83f6-fd1ee450ff15.png)

mrDIMAS:
Interesting bug, it is definitely should be fixed asap. Thanks for the report.

lvella:
I just opened the editor and it stood there, doing nothing but consuming 50% of the CPU. Same thing both in release and debug mode.

mrDIMAS:
Fixed in #520.

PR #520, opened 2023-09-05
Optimize PC resources consumption in the editor

mrDIMAS:
This PR significantly reduces PC resources consumption by the editor by stopping its update loop when there's in messages from the OS or there's no user actions. Preview mode keeps the editor updating every frame, which is acceptable since it is not used very often.


mrDIMAS:
Fixes #360.

PR #529, opened 2023-09-14
Nitpicks

adamnemecek:


mrDIMAS:
The CI seems to fail in some const fns, could you please fix those? Also this PR is quite large and will take some time to review properly.

adamnemecek:
I’ll fix those. It’s quite manageable when you look at the single commits in separation as most of them were find and replace.

PR #549, opened 2023-10-27
Feature: ResourceIo system

### LINKED ISSUE #488: FEATURE PROPOSAL: Pluggable VFS for loading assets
We need a pluggable virtual file system for assets. With that, we can implement any of the archive formats for loading assets.

jacobtread:
## Description

> Sorry in advance for the large commit

> **Warning**
> This is a breaking change, any custom resource loaders will need to be extended as I modified the arguments for load. Things that read files such as DataSource require an additional argument for the resource_io to use (I've updated the examples, tests, and editor to just use FsResourceIo)

This is basically a virtual file system implementation

I have swapped out a majority of the resource handling code (I think there's still some code that's using WalkDir however thats in the editor code only so I don't think its nessicary to port it) with an interface that I have added to the ResourceManager named ResourceIo, this is a similar abstraction to what the Bevy engine uses and abstracts the file reading and exist checking operations away from the std::fs types.

The interface provides `load_file` which reads the entire bytes of the file, `file_reader` which opens the file for streamed reading (The fallback for android and wasm just uses a cursor wrapped around the bytes from `load_file`) and `exists` which returns whether the path exists or not, `walk_directory` which provides directory walking (Uses walkdir but is no-op for android and wasm as im not sure if that can be done easily on those platforms) , `is_file` which checks if the provided path is a file, and `is_dir` which checks if the provided path is a directory (Defaulting this to false on wasm as I don't know if that can be checked reliably) `read_directory` which provides a list of paths within a directory (currently no-op for android and wasm but an android impl is likely possible just will require some extra work)

I have created a default implementation called FsResourceIo which is just an abstraction over your existing io functions from fyrox_core::io along with some new ones that I've added
image

I went through and replaced most of the existing usages of the functions from fyrox_core::io with usages of the abstraction

## Reason

This abstraction allows resources to be loaded from custom and packed file formats rather than having resources only and always loaded from file system paths. My use case for this is the game uses a custom packed file format to store all of its assets so I can use this feature to seek and read within the packed file format without having to extract it. It also enables the usage of zip files for storing assets if someone wants to make a ResourceIo for that purpose.

Below is my example for the custom packed asset format ResourceIo:
image

Then swapping the default resource_io:
image

## Changes
- Added ResourceIo trait for providing IO operations for resource loading
  - With default implementation of FsResourceIo
- Added resource_io to the resource manager state
- Added helper function on resource manager for getting an Arc to the current resource_io
- Added setter to the resource manager state for setting the current resource_io
- Modified ResourceLoader load function to add the resource_io as an argument
- Replaced the previous usages of fyrox_core::io with usages of the resource_io abstraction
- Added FileReader trait to be used as a reader abstraction for readers provided by the ResourceIo 
- Swapped compile time wasm checking in DataSource with Boxed dynamic usage of FileReader provided by ResourceIO (already has the wasm handling built in as the default impl)
- Added additional IO functions to fyrox_core::io 

## Related issues
- Closes #488 
  - This change allows implementing and loading from archive file formats

## Not implemented

I have not implemented the system fully into the editor, the editor just uses FsResourceIo for all the operations that require it as it will likely require some extra ui and such for including other formats into the editor (Not sure how youd go about that)

### Previous

```rust
#[derive(Eq, PartialEq, Debug)]
struct MyResourceLoader;

impl ResourceLoader for MyResourceLoader {
    fn extensions(&self) -> &[&str] {
        &[]
    }

    fn data_type_uuid(&self) -> Uuid {
        Default::default()
    }

    fn load(
        &self,
        resource: UntypedResource,
        event_broadcaster: ResourceEventBroadcaster,
        reload: bool,
    ) -> BoxedLoaderFuture {
        Box::pin(async move {
            let bytes = fyrox_core::io::load_file(resource.path().as_ref()).await.unwrap();
            // ... do something with bytes
            Ok(())
        });
    }
}
```

### New

```rust
#[derive(Eq, PartialEq, Debug)]
struct MyResourceLoader;

impl ResourceLoader for MyResourceLoader {
    fn extensions(&self) -> &[&str] {
        &[]
    }

    fn data_type_uuid(&self) -> Uuid {
        Default::default()
    }

    fn load(
        &self,
        resource: UntypedResource,
        event_broadcaster: ResourceEventBroadcaster,
        reload: bool,
        io: Arc<dyn ResourceIo>,
    ) -> BoxedLoaderFuture {
        Box::pin(async move {
            let bytes = io.load_file(resource.path().as_ref()).await.unwrap();
            // ... do something with bytes
            Ok(())
        });
    }
}
```

### DIFF
--- a/editor/src/asset/inspector/handlers/model.rs
+++ b/editor/src/asset/inspector/handlers/model.rs
@@ -1,6 +1,7 @@
 use crate::asset::inspector::handlers::ImportOptionsHandler;
 use fyrox::{
     asset::{
+        io::FsResourceIo,
         manager::ResourceManager,
         options::{try_get_import_settings, ImportOptions},
     },
@@ -19,7 +20,8 @@ impl ModelImportOptionsHandler {
     pub fn new(resource_path: &Path) -> Self {
         Self {
             resource_path: resource_path.to_owned(),
-            options: block_on(try_get_import_settings(resource_path)).unwrap_or_default(),
+            options: block_on(try_get_import_settings(resource_path, &FsResourceIo))
+                .unwrap_or_default(),
         }
     }
 }
@@ -33,7 +35,8 @@ impl ImportOptionsHandler for ModelImportOptionsHandler {
     }
 
     fn revert(&mut self) {
-        self.options = block_on(try_get_import_settings(&self.resource_path)).unwrap_or_default();
+        self.options = block_on(try_get_import_settings(&self.resource_path, &FsResourceIo))
+            .unwrap_or_default();
     }
 
     fn value(&self) -> &dyn Reflect {
--- a/editor/src/asset/inspector/handlers/sound.rs
+++ b/editor/src/asset/inspector/handlers/sound.rs
@@ -1,6 +1,7 @@
 use crate::asset::inspector::handlers::ImportOptionsHandler;
 use fyrox::{
     asset::{
+        io::FsResourceIo,
         manager::ResourceManager,
         options::{try_get_import_settings, ImportOptions},
     },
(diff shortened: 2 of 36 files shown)


mrDIMAS:
Very nice PR, thank you! The editor could live without VFS for now, so no problems with that. It could be changed later if needed.

Barugon:
Nice!

PR #554, opened 2023-10-30
Feature: UI palette

### LINKED ISSUE #550: Feature: Widget color palette
## Description

> This is a feature which i'm looking to implement into fyrox ui rather than a feature request

Widget palette structure for Widgets that defines a set of colors for the widget to use with different states (i.e Hover, Normal, Pressed, Selected) similar to the brush fields on the Decorator to allow all widgets to react to different states with colors (i.e the border of a widget can be transparent then given a color when the widget is hovered) or some other solution making the widgets more styleable 

I will begin work on figuring out what a good way to do this would be and I will update this issue with my progress on it

jacobtread:
## Description

This feature adds a widget palette system onto widgets themselves which defines the brushes to use for the normal and hover states (On all widgets).

This enable things like the border color changing when a button is hovered as opposed to this only being supported by the background color for Decorator.

This should be backward compatible with the existing `with_background`/`with_hover_brush`/`with_normal_brush` system thats already in place as I modified those functions to update the palette instead.

## Usage

The `WidgetBuilder` structure now has a `palette` field and a `with_palette` function which takes in a `WidgetPaletteBuilder` which you can use to define the colors for the different background and foreground states:

image

There are also `with_background` and `with_foreground` functions on the palette builder to set both the hovered and normal colors (If a hovered color is not picked but a normal color is then when the palette is built it will use the normal color for the hovered color)

image

## In-game usage

Below is usage of this feature to make the border color of buttons transparent until they are hovered:

https://github.com/FyroxEngine/Fyrox/assets/33708767/a8d6a5d8-9774-4724-8e57-4532630ad1fd

## Related Issue
- Closes #550 (Tracking issue for this implementation)


fyrox-ui/src/decorator.rs
```diff
@@ -290,14 +251,25 @@ impl DecoratorBuilder {
     }
 
     /// Sets a desired brush for `Normal` state.
+    ///
+    /// ## Backward compatablity
```
mrDIMAS, line 255: typo `compatablity`

fyrox-ui/src/decorator.rs
```diff
@@ -207,6 +181,14 @@ impl Control for Decorator {
                         ));
                     }
                 }
+
+                // Backward compatablity (Updates colors on underlying palette)
```
mrDIMAS, line 185: typo `compatablity`

fyrox-ui/src/decorator.rs
```diff
@@ -290,14 +251,25 @@ impl DecoratorBuilder {
...
+    /// a `WidgetPaletteBuilder` on the underlying widget
     pub fn with_normal_brush(mut self, brush: Brush) -> Self {
-        self.normal_brush = brush;
+        self.border_builder.widget_builder.palette.background_normal = Some(brush);
         self
     }
 
     /// Sets a desired brush for `Hovered` state.
+    ///
+    /// ## Backward compatablity
```
mrDIMAS, line 266: typo `compatablity`

mrDIMAS (changes requested):

fyrox-ui/src/widget.rs
```diff
@@ -550,6 +559,166 @@ impl WidgetMessage {
...
+    ForegroundNormal(Brush),
+    /// Set the palette hover foreground brush.
+    ForegroundHover(Brush),
+}
+
+impl WidgetPalette {
+    /// Updates the background and foreground color of the provided
+    /// node to be the normal colors from this palette
+    pub fn set_normal(&self, handle: Handle<UiNode>, ui: &mut UserInterface) {
+        ui.send_message(WidgetMessage::background(
```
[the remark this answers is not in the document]
jacobtread: From looking at the code I've just realized there isn't actually any need for me to be sending messages here as it can just have access to the widget itself instead and update it from there, the new code:
image

mrDIMAS:
Have you checked the editor with your changes? It is works as before?

jacobtread:
I have not checked the editor i will have to get a game setup with one and will check that shortly

jacobtread:
Ah the performance in the editor for the editor tabs is heavily degraded from the nested widgets and having to do `has_descendant` to check descendants for hovering, looks like the custom widget approach you mentioned in #550 is definitely the better way to go.

Looks like i'll close this pull request as it doesn't seem like the viable approach unless you've got any ideas for making it work well?

jacobtread:
After switching over to custom widgets they are definitely the way to go, closing this pr

PR #560, opened 2023-11-05
Added a small new feature in absm editor

Riddhiman007:
Connecting transitions from one state to another is a boring task, especially when many states are there in the absm editor.
Here's a preview:

https://github.com/FyroxEngine/Fyrox/assets/91105040/5361e196-95ff-436b-8ac9-8daeab10fb36

P. S: You've to click the state graph to confirm the transition (I can't figure out how to solve that).

### DIFF
--- a/src/material/shader/mod.rs
+++ b/src/material/shader/mod.rs
@@ -146,21 +146,25 @@
 //! - PointShadow - A pass that emits distance from a fragment to a point light, later this depth
 //! map will be used to render shadows.
 //!
-//! # Built-in variables
+//! ## Built-in properties
 //!
-//! There are number of build-in variables that Fyrox pass to each shader automatically:
+//! There are number of built-in properties, that Fyrox will try to assign automatically if they're defined
+//! in your shader:
 //!
-//! | Name                      | Type            | Description
-//! |---------------------------|-----------------|--------------------------------------------
-//! | fyrox_worldMatrix          | `Matrix4`       | Local-to-world transformation.
-//! | fyrox_worldViewProjection  | `Matrix4`       | Local-to-clip-space transform.
-//! | fyrox_boneMatrices         | `[Matrix4; 60]` | Array of bone matrices.
-//! | fyrox_useSkeletalAnimation | `Vector3`       | Whether skinned meshes is rendering or not.
-//! | fyrox_cameraPosition       | `Vector3`       | Position of the camera.
-//! | fyrox_usePOM               | `bool`          | Whether to use parallax mapping or not.
-//! | fyrox_lightPosition        | `Vector3`       | Light position.
+//! | Name                       | Type         | Description                                                                                                       |
+//! |----------------------------|--------------|-------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


editor/src/absm/canvas.rs
```diff
@@ -384,12 +398,30 @@ impl Control for AbsmCanvas {
...
                 );
             }
+            Mode::CreateTransitionToAllNodes {
+                source_pos,
+                dests_pos,
+                ..
+            } => {
+                for dest_pos in dests_pos {
+                    if dest_pos != source_pos {
+                        // Log::info(format!("State Position: {:?}", dest_pos));
```
mrDIMAS, line 414: Remove commented code please.

editor/src/absm/state_graph/mod.rs
```diff
@@ -210,6 +210,22 @@ impl StateGraphViewer {
...
                     }
+                    AbsmCanvasMessage::CommitTransitionToAllNodes {
+                        source_node,
+                        dest_nodes,
+                    } => {
+                        if message.direction() == MessageDirection::FromWidget {
+                            let source = fetch_state_node_model_handle(*source_node, ui);
+                            for dest_node in dest_nodes {
+                                let dest = fetch_state_node_model_handle(*dest_node, ui);
+                                sender.do_scene_command(AddTransitionCommand::new(
```
mrDIMAS, line 221: Here you create a separate command per each transition, and if you try to undo this in the editor, you'll see that each of your transitions will be removed one-by-one. You should use command group instead. For example, when you dragging multiple nodes, the editor creates a group command for all nodes (https://github.com/FyroxEngine/Fyrox/blob/master/editor/src/absm/state_graph/mod.rs#L159). You could use this as an example.

mrDIMAS (changes requested):
Nice functionality, thanks. Please fix all the comments and I'll merge the PR.

editor/src/error.rs
```diff
@@ -0,0 +1,29 @@
...
+    core::{algebra::Vector2, color::Color},
+    gui::{
+        brush::Brush, text::TextBuilder, widget::WidgetBuilder, window::WindowBuilder,
+        UserInterface,
+    },
+};
+
+pub struct ErrorViewer;
+
+impl ErrorViewer {
```
mrDIMAS, line 11: Just print an error into the log via `Log::err("message")` and it will automagically appear in the editor's log.

editor/src/interaction/terrain.rs
```diff
@@ -116,11 +117,17 @@ impl BrushGizmo {
     }
 }
 
-fn copy_layer_masks(terrain: &Terrain, layer: usize) -> Vec<Vec<u8>> {
+fn copy_layer_masks(terrain: &Terrain, layer: usize, ui: &mut UserInterface) -> Vec<Vec<u8>> {
```
mrDIMAS, line 120: This should be in a separate pull request, because it does not have anything in common with ABSM editor.

mrDIMAS (changes requested):
Please move terrain-related code into a separate PR.

PR #563, opened 2023-11-10
Enhancements in the absm editor

Riddhiman007:
This is similar to my previous pr but the problem that @mrDIMAS found were solved.

**Enhancements made in the ABSM editor:**
- There is a menu added in the node context menu of the absm editor named **Create transition to all nodes** which will create transition from the source node to all states.
- In the state graph viewer similar menu is added which will connect all nodes.

Hope that this will save time while blending animations.
If any problem is found please comment me.

### DIFF
--- a/editor/src/absm/state_graph/context.rs
+++ b/editor/src/absm/state_graph/context.rs
@@ -1,8 +1,8 @@
 use crate::{
     absm::{
-        canvas::{AbsmCanvasMessage, Mode},
+        canvas::{AbsmCanvas, AbsmCanvasMessage, Mode},
         command::{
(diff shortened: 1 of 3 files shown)


editor/src/absm/state_graph/context.rs
```diff
@@ -149,6 +201,7 @@ impl NodeContextMenu {
         absm_node_handle: Handle<Node>,
         absm_node: &AnimationBlendingStateMachine,
         layer_index: usize,
+        // canvas: Handle<UiNode>,
```
mrDIMAS, line 204: Why is this commented?

mrDIMAS (changes requested):

Issue #559, opened 2023-11-02
Selecting and Deselecting Objects Cause Memory Increasing (Leaking?)


IceGuye:
There are a few occasions which may cause memory increasing (or leaking?), such as selecting and deselecting objects.

System: Feodra Linux 38 with Gnome X11.
Version: Newest git master.
Reproduce:
1) Use fyrox-template to create a brand new 3D project.
2) Change the cargo.toml to the git master.
3) Start the editor with `cargo run --package editor --release`.
4) Select and deselect some object, such as the default cude, light, and camera.
5) Check the system monitor to see the memory increasing.

See the video for details.

https://github.com/FyroxEngine/Fyrox/assets/7291463/7b8636d1-48b1-4a03-bc7c-ea1fb0972f87

mrDIMAS:
~~Works on my PC~~ I can't reproduce this on Windows, I selected and deselected various objects dozens of time and memory usage was increased on just 1 Mb. Which is expected, because selection/deselection creates a bunch of commands.

LordCocoNut:
I think this issue could be closed, according to [discord conv](https://discord.com/channels/756573453561102427/941678932791529523/1178054198361870387)

PR #569, opened 2023-11-25
github #559 fix for linux memory leaks

LordCocoNut:
- serialization of settings was replaced with file::write_all and ron::ser::to_string_pretty to prevent memory leaks on linux
- this should also make editor run smoother and limit few lags eg. during moving a mesh around

### DIFF
--- a/editor/src/settings/mod.rs
+++ b/editor/src/settings/mod.rs
@@ -37,6 +37,7 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     fs::File,
+    io::Write,
     ops::{Deref, DerefMut},
     path::PathBuf,
     rc::Rc,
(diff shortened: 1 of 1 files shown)


editor/src/settings/mod.rs
```diff
@@ -165,9 +166,11 @@ impl SettingsData {
     }
 
     fn save(&mut self) -> Result<(), SettingsError> {
-        let file = File::create(Self::full_path())?;
+        let mut file = File::open(Self::full_path())?;
```
mrDIMAS, line 169: This will fail if the settings file does not exist. It should be `File::create` as before.

mrDIMAS (changes requested):
One small comment to fix and I'll merge the PR.

PR #599, opened 2024-02-05
Open script in external editor: initial support

Voronar:
Basic implementation, tested only in VSCode on MacOS. For a better experience your game project should be already opened in the selected external editor.

# How to use

It works automagically.

# Settings
image

# Inspector
image


fyrox-core/src/script.rs
```diff
@@ -0,0 +1,11 @@
+pub use fyrox_core_derive::ScriptSourcePathProvider;
+
+pub mod prelude {
+    pub use super::ScriptSourcePathProvider;
+}
+
+/// Script source path provider
+pub trait ScriptSourcePathProvider: Sized {
+    /// Script source path to open it from Editor
+    fn script_source_path() -> &'static str;
```
mrDIMAS, line 10: It is probably worth putting this functionality to `Reflect` trait and call it `source_path()`. This might sound counter-intuitive, but, for instance, Java supports this kind of functionality via `CodeSource`. This way you don't need to introduce the new trait and the procmacro for it. Also `add_with_source_path` is not needed in this approach, all previously made scripts will automagically know about their source path.

PR #624, opened 2024-04-04
glTF Model Loader

b-guild:
This is a loader for glTF files that looks for .gltf and .glb file extensions and loads models using the gltf crate.

It adds three features to fyrox, fyrox-impl, and fyrox-dylib:

- gltf: Enable the dependency on the gltf crate and enable the loader which depends upon it, including the gltf module in fyrox_impl::resource.
- gltf_blend_shapes: Enable glTF blend shapes, which the glTF format calls "morph targets." Loading morph targets requires enabling the "extras" feature in the gltf crate, and this feature controls whether that is enabled and whether the loader looks for blend shapes.
- mesh_analysis: Causes the glTF loader to search every loaded triangle for zero-length edges and log an error if they are found.

### DIFF
--- a/editor/src/export.rs
+++ b/editor/src/export.rs
@@ -650,7 +650,30 @@ fn export(export_options: ExportOptions, cancel_flag: Arc<AtomicBool>) -> Result
                     Log::verify(open::that_detached("http://127.0.0.1:4000"));
                 }
                 TargetPlatform::Android => {
-                    Log::err("Unimplemented yet!");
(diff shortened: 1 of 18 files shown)


mrDIMAS:
Insanely cool PR, thank you! It will take some time to read it.

mrDIMAS:
Everything looks ok from simple reading of the PR (took me a few hours to do 😅), there are tiny quirks, but they're insignificant.

PR #627, opened 2024-04-08
TextBox ergonomic adjustments.

b-guild:
I have made countless tiny adjustments to TextBox in order to make it more comfortable for users. I did not like how it felt to use text boxes in the UI before, and now they feel very nice. Try it and see if you agree that they are better this way. Both mouse and keyboard interaction have changed. Ctrl-X now does cut, much like Ctrl-C does copy and Ctrl-V does paste. Mouse clicks are detected in a *far* more forgiving way, even allowing mouse drags to stray outside of the text box.

In order to make it easier to get TextBox working the way I want, I have made several internal adjustments to text_box.rs and formatted_text.rs.

* I have expanded the capabilities of FormattedText to include several useful tools to simplify text_box.rs. FormattedText now knows about Position, can determine the start and end Position of each line, and can translate between local points and Positions. These are all capabilities that require no knowledge of text being edited, so it seems reasonable to reduce the burden of responsibility from the very heavily burdened text_box.rs.
* I have created a new sub-module within `formatted_text` called `textwrapper`. It is responsible for splitting the chars of FormattedText into lines based on advance distances provided by FormattedText. Properly wrapping text is famously challenging to implement perfectly, so for my own peace of mind I needed to separate that task from the rest, and separate the three different kinds of wrapping into three separate functions. Originally all different kinds of text wrapping was handled within a single loop within a large function within FormattedText, but the complexity of that boggled my mind. The `textwrapper` module is entirely self-contained, knowing nothing of FormattedText or Fyrox.


mrDIMAS:
I tried the changes locally and it works smoothly and feels nice, great job. I just need to read the rest (about a half) of the code.

Issue #199, opened 2021-10-08
Scrollbar has fixed size and sometimes is so big that it can't be dragged.


martin-t:
In the screenshot you can see 4 entries but there are actually 6. The widget is however so small that dragging the scrollbar with mouse is impossible - i have to click the down arrow (or... thinking about it now... use the mouse wheel but mine is broken so it didn't occur me until now ;) ).

The underlying issue is that the scrollbar has a fixed size but increasing the side of this widget would be a nice thing too.

![2021-10-08-182944_418x282_scrot](https://user-images.githubusercontent.com/4079823/136591684-5156e154-5c45-40ca-bbc9-6ee3edfa1a60.png)

shyamjayakannan:
@mrDIMAS will decreasing the minimum size here help?

https://github.com/FyroxEngine/Fyrox/blob/7de5a3107da2dfab180fd23f4c79b2542b466d70/fyrox-ui/src/scroll_bar.rs#L548-L555

mrDIMAS:
I think the size of the `indicator` should be adaptive to the actual size of the content of a `ScrollViewer`. This issue is a bit more complex than just decreasing the size, simply because currently the size is fixed and decreasing it would result in quite unusable scroll bar.

PR #628, opened 2024-04-08
added code to make indicator size adaptive

shyamjayakannan:
#199 

Added code to make the `indicator` (thumb) size of the scrollbar adaptive to the `ScrollViewer`. Behavior shown below:

https://github.com/FyroxEngine/Fyrox/assets/120047549/09f05263-d381-44eb-bd9a-a2d5a93085e4

I hope the behavior is as expected. But there is an issue:
It seems to be working fine when the container is extended, but when it is shrunk below a certain size, the indicator's movement gets offset by a certain distance (can be seen in the video where the indicator doesn't touch the top arrow and extends below the bottom arrow)

Any help with this would be greatly appreciated. I hope this helps

Also, I have set the minimum size of the indicator to 15. Please let me know if this should be changed or handled differently.

### DIFF
--- a/fyrox-ui/src/scroll_bar.rs
+++ b/fyrox-ui/src/scroll_bar.rs
@@ -39,6 +39,9 @@ pub enum ScrollBarMessage {
     /// Used to indicate that the max value of the scroll bar has changed ([`MessageDirection::FromWidget`]) or to set a
     /// new max value (with [`MessageDirection::ToWidget`].
     MaxValue(f32),
+    /// Used to set the size of the indicator(thumb) adaptively, according to the relative sizes of the container and the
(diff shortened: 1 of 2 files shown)


mrDIMAS:
The code looks nice, great job. The issue you described looks to be related to this https://github.com/FyroxEngine/Fyrox/blob/b45c90960c1292bd693a544eb7f10cf15ec29c4e/fyrox-ui/src/scroll_bar.rs#L588-L595 min size settings. I think it should be set to 15 too and the issue should disappear.

shyamjayakannan:
Thank you so much @mrDIMAS . I have made the changes and everything looks fine.

PR #632, opened 2024-04-14
Prevent Inspector properties from disappearing when expander is closed.

b-guild:
Many inspector properties include expander widgets to allow those editors to be hidden, but often closing those expanders will cause the entire property to disappear, so it is impossible to open the expander. This is an attempt to solve that problem.

The source of the problem was that InheritablePropertyEditor would re-send messages from its inner editor but with the destination changed to be the InheritablePropertyEditor, and this included re-sending messages that were sent to the inner editor, like visibility messages. So when the expander told the inner editor to become invisible, the InheritablePropertyEditor repeated that message to told itself to also become invisible.

I have hopefully resolved that problem by having the InheritablePropertyEditor not re-send ToWidget messages. Doing this *seems* to have no unintended side-effects. If InheritablePropertyEditor were to re-send ToWidget messages, it seems that the proper way to do so would be by re-sending its own messages redirected toward the inner editor, but that is probably not necessary.


mrDIMAS:
Very cool, I've spent few days trying to fix this issue back in the day, but then just gave up.

PR #648, opened 2024-05-10
Changing AABB validity to include zero-size dimensions to allow camera fitting to work with 2D objects.

b-guild:
I noticed that camera fitting tended to fail with some objects in my scene and I tracked the problem down to how the fitting algorithm excludes certain AABBs as "invalid" when the only problem with them is that they have zero length in one dimension. I think what validity is *really* supposed to check for that an AABB is not negative in any of its dimensions, so I changed the test for positive into a test for not-negative.

### DIFF
--- a/editor/src/camera/mod.rs
+++ b/editor/src/camera/mod.rs
@@ -187,7 +187,7 @@ impl CameraController {
         let mut aabb = AxisAlignedBoundingBox::default();
         for descendant in scene.graph.traverse_iter(handle) {
             let descendant_aabb = descendant.local_bounding_box();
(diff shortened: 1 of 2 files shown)


mrDIMAS (changes requested):
The tests for AABB validity seem to fail.

b-guild:
I think the problem was with the test, not with `is_valid`. Surely zero-size AABBs should be considered valid. We have `is_degenerate` to test for zero-size AABBs, and we have `is_invalid_or_degenerate` which would be a completely redundant function if all degenerate AABBs are invalid.

Issue #630, opened 2024-04-11
Unsound implementation of `transmute_vec_as_bytes` in `fyrox-core`


shinmao:
Hi, we are researchers from [SunLab](https://sunlab-gmu.github.io/). We consider the safe function `transmute_vec_as_bytes` includes the unsound implementation.

### `transmute_vec_as_bytes`
https://github.com/FyroxEngine/Fyrox/blob/3b03ea717d201be24dac98cc9e8ebf812808e898/fyrox-core/src/lib.rs#L325-L331
When casting type to `u8` slice, we also need to make sure whether the type won't contain padding bytes. Otherwise, it could lead to uninitialized memory access or break the reliability of program. Based on the usages of the function, we consider the generic type `T` should also implement `Pod` trait.

### PoC
```rust
use fyrox_core::transmute_vec_as_bytes;

#[derive(Copy, Clone)]
struct Pad {
    a: u8,
    b: u32,
    c: u8
}

fn main() {
    let pd = Pad { a: 0x1, b: 0x2, c: 0x3 };
    let mut v = Vec::new();
    v.push(pd);
    let fv = transmute_vec_as_bytes(v);
    println!("{:?}", fv);
}
```
In the program above, we passed the `struct` that might contain padding bytes as the generic type `T`. First, when we run with miri, we can get the following results:
```
error: Undefined Behavior: using uninitialized data, but this operation requires initialized memory
   --> /home/rafael/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/num.rs:471:5
    |
471 | /     impl_Display!(
472 | |         i8, u8, i16, u16, i32, u32, i64, u64, usize, isize
473 | |             as u64 via to_u64 named fmt_u64
474 | |     );
    | |_____^ using uninitialized data, but this operation requires initialized memory
```
In the library, we found that this function uses `Vec` as input as internal usages, which will not trigger the bug. However, since this function is provided as public safe API, we should consider the constraints on the generic type `T`. (It is required also because the function name is `transmute_"vec"_as_bytes` :)

### The consequence of UB
We could find that the results of `fv` break the reliability of program under different environment.
Compiled with `x86_64`, the results would be:
```
[2, 0, 0, 0, 1, 3, 0, 0]
```
While compiled with `x86`, the results would be:
```
[2, 0, 0, 0, 1, 3, 233, 247]
```
Take the following usages for example,
[...]

PR #662, opened 2024-06-30
#630 transmute_vec_as_bytes potential fix

NathanaelG1:
Hey all new here but believe this may fix the issue raised in #630. Added some comments on the issue thread including a playground which shows the result of the old function versus the new function. I have taken an approach to only account for the specific types which actually are used in the codebase (f32 and usize) as there were a lot of issues dealing with a totally generic type. I understand this adds a limitation that was previously not there, if this is not a good approach let me know and I can do some reformatting. 

I have added two unit tests for transmute_vec_as_bytes which makes sure the results do not include any random values from the padding. 

I removed the need for an unsafe block because the types are known to the function, and did not add any dependencies to the project.

All unit tests are passing, and running my modified tutorials from the RPG and FPS on the documentation seem to run the same as on the master branch of the main fork. This is my first PR here if there's anything I can do to improve it let me know!


mrDIMAS:
Hi, this looks like a partial solution. May I ask you, could you please try using `bytemuck` crate for this case? Afaik all you need to do is to add `Pod` trait bound to `T` and this should automagically fix the issue.

NathanaelG1:
@mrDIMAS 

I have added bytemuck and added the pod trait to T, on compiling there is an error that Pod should be implemented for BlendShapesContainer::from_lists::VertexData. When I implement it there using the derive macro there are 3 more errors for every Vec3 declared in this impl block. 

Using bytemuck may require a lot more changes to the code I haven't gotten through them all yet, minimally bytemuck will need to be added to fyrox-core and fyrox-impl and a pod wrapper with implementations of Pod and Zeroable to be added to Vertex Data. If this approach sounds good I'll continue.

NathanaelG1:
[@mrDIMAS](https://github.com/mrDIMAS)

I have added `bytemuck` and applied the `Pod` trait to `T`. During compilation, an error indicated that `Pod` should be implemented for `BlendShapesContainer::from_lists::VertexData`. When I attempted to derive `Pod` for `VertexData`, additional errors surfaced for each `Vec3` declared in the implementation block.

Using `bytemuck` might require significant changes to the codebase. At a minimum, `bytemuck` would need to be added to both `fyrox-core` and `fyrox-impl`, and a `Pod` wrapper with implementations of `Pod` and `Zeroable` would need to be created for `VertexData`.

If this approach sounds acceptable, I'll proceed with the necessary changes. 

errors: 
   --> fyrox-impl/src/scene/mesh/surface.rs:170:21
    |
170 |         let bytes = crate::core::transmute_vec_as_bytes(vertex_data);
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type of the type parameter `Pod` declared on the function `transmute_vec_as_bytes`
    |

   --> fyrox-impl/src/scene/terrain/mod.rs:116:9
    |
116 |         crate::core::transmute_vec_as_bytes(height_map),
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type of the type parameter `Pod` declared on the function `transmute_vec_as_bytes`
    |

mrDIMAS:
Yes, you can proceed. As for `nalgebra` items (Vectors, etc) - you can add `bytemuck` feature to `nalgebra` crate, and it will implement the required traits.

NathanaelG1:
Made the suggested changes and left the unit tests.

fyrox-impl/src/scene/mesh/surface.rs
```diff
@@ -94,18 +95,45 @@ pub struct InputBlendShapeData {
...
+        }
+    }
+}
+
+impl From<PodVector3> for Vector3<f16> {
+    fn from(pv: PodVector3) -> Self {
+        Vector3::new(pv.x, pv.y, pv.z)
+    }
+}
+
```
mrDIMAS, line 124: `half` crate has `bytemuck` feature as well, so this `PodVector3` can be replaced by simply enabling the `bytemuck` feature.

mrDIMAS (changes requested):

mrDIMAS (approved):

Issue #586, opened 2024-01-17
Add an ability to rotate the editor camera using scene gizmo


mrDIMAS:
Add an ability to rotate the editor camera using scene gizmo:

![scene gizmo](https://github.com/FyroxEngine/Fyrox/assets/2785135/4fd16084-775a-45e1-b24a-27c4ece1e845)

Rotation should happen if a user clicks on the gizmo and starts dragging the mouse while holding the button. Ordinary clicks should switch views as before. Scene gizmo is located in `editor/src/scene_viewer/gizmo.rs`

Fedron:
I would like to take this one on if possible?

mrDIMAS:
Hi, yeah, this issue seems to be quite easy to fix, so you can try it.

PR #669, opened 2024-07-12
#586 Add an ability to rotate the editor camera using scene gizmo

NathanaelG1:
I have updated the scene_viewer module to listen for a left click and follow the mouse while syncing with the main scene camera. This seems to work well however occasionally the MouseUp event is not heard and the property is_left_mouse_clicked remains true unless you click again. 

I wanted to get some feedback at this stage. Comparing this gizmo functionality to what is in Blender, when the MouseOver is detected on the gizmo a sphere appears indicating the perspective can be changed by clicking. 

image

Would it be good to have a similar indicator for this by adding a shape to the gizmo that is visible on mouse over?


NathanaelG1:
https://drive.google.com/file/d/1pR5zShXht5kuMJEinhSUfRKffGdmfMAl/view?usp=drive_link <-- public link to video showing gizmo functionality.

editor/src/scene_viewer/gizmo.rs
```diff
@@ -206,6 +210,13 @@ impl SceneGizmo {
             pos_z,
             neg_z,
             center,
+            is_left_mouse_pressed: false,
+            last_mouse_position: None,
+            camera_rotation: CameraRotation {
```
mrDIMAS, line 215: This is not needed, all you need to do is to pass a mutable reference to the current camera controller instance and change its `pitch` and `yaw` fields accordingly to the mouse movement - horizontal movement should change the yaw, and the vertical movement should change the pitch.

mrDIMAS (changes requested):

editor/src/scene_viewer/gizmo.rs
```diff
@@ -297,6 +302,11 @@ impl SceneGizmo {
                 },
             )
         }
+
+        if let Some(camera_controller) = camera_controller {
+            camera_controller.yaw = pos.x * 0.1;
```
mrDIMAS, line 307: You need to use relative values, not absolute position here. Otherwise the camera rotation will jump unpredictably when you start rotating the camera. So basically you need to save click position somewhere and calculate `let delta = pos - self.click_position` and use `delta` instead of `pos` in your code.

editor/src/scene_viewer/mod.rs
```diff
@@ -827,14 +826,21 @@ impl SceneViewer {
...
                                 let rel_pos = pos
                                     - engine
                                         .user_interfaces
                                         .first()
                                         .node(self.scene_gizmo_image)
                                         .screen_position();
-                                self.scene_gizmo.on_mouse_move(rel_pos, engine);
+                                self.scene_gizmo.on_mouse_move(rel_pos, engine, None);
+                                if state.left == ButtonState::Pressed {
+                                    self.scene_gizmo.on_mouse_move(
```
mrDIMAS, line 838: Instead of double call just change signature of `on_mouse_move` to accept mouse button and its state and check for pressed/unpressed in `on_mouse_move` itself.

mrDIMAS (changes requested):

editor/src/scene_viewer/gizmo.rs
NathanaelG1: Okay interesting, I was using delta originally but when doing some reformatting I noticed it seemed pretty smooth just going off mouse movement. Didn't notice I could get the click position 👍

NathanaelG1:
@mrDIMAS I have made the requested changes. I have noticed without locking the camera_controller.z_offset the camera does a sort of orbital motion around its current perspective. I found this kind of nice for navigating a scene and getting different angles as holding down right mouse button does the in place look around. 

Should the basic position of the camera be locked to prevent this? I have implemented it both ways just wondering what would be best. 

Attached a video showing the orbital rotation
 

https://github.com/user-attachments/assets/51a6cf6e-4833-406d-9c65-fadab6761074

mrDIMAS:
I think orbital movement is preferred in this case, I just checked Blender and it does orbital movement as well.

editor/src/scene_viewer/gizmo.rs
```diff
@@ -297,9 +306,17 @@ impl SceneGizmo {
                 },
             )
         }
+        if left_mouse_button == ButtonState::Pressed {
+            let delta = pos - self.last_click_position;
+            if delta.magnitude() > 0.1 {
+                camera_controller.yaw = delta.x * 0.1;
```
mrDIMAS, line 312: You also need to store initial values of yaw and pitch (at the click moment) and basically do this: `camera_controller.yaw = self.initial_yaw + delta.x * 0.1`. Otherwise you'll get jumping camera, which is already the case on the video you've attached.

mrDIMAS (changes requested):

editor/src/scene_viewer/gizmo.rs
NathanaelG1: I am now storing the yaw and pitch when a left click technically on the scene is pressed but it also updates on clicking the gizmo. If I should place it further down I can but this seemed to make it better. 

I think another issue potentially is not having some element to grab onto so when trying to rotate the gizmo via click it first picks a node to switch perspective. 

I could place this functionality behind some kind of mesh sphere to help with this or if you have a suggestion there. Thanks for your help i've learned so much about the editor especially on this one 🙏

mrDIMAS: All you need to do is to move interaction with arrows to mouse up event (instead of mouse down like it is now). It is located here - https://github.com/FyroxEngine/Fyrox/blob/master/editor/src/scene_viewer/mod.rs#L785, so you basically need to change `MouseDown` to `MouseUp` and add new match branch with `MouseDown` and put there rotation handling (its initiation, to be precise). Also, to prevent triggering of arrows on MouseUp when you end dragging, you may also add `is_dragging: bool` flag and raise it when you start dragging (when movement delta is large enough). Check this flag in `on_click` method and return if it is raised.

mrDIMAS (changes requested):

editor/src/scene_viewer/gizmo.rs
```diff
@@ -49,6 +50,10 @@ pub struct SceneGizmo {
     pub pos_z: Handle<Node>,
     pub neg_z: Handle<Node>,
     pub center: Handle<Node>,
+    pub first_yaw: f32,
+    pub first_pitch: f32,
+    pub is_dragging: bool,
+    pub first_click_pos: Vector2<f32>,
```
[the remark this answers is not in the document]
NathanaelG1: I was thinking of grouping them in a struct called SceneGizmoOrbital. Something like
``` rust 
struct SceneGizmoOrbital {
    pub initial_click_pos: Vector2<f32>,
    pub is_dragging: bool,
    pub gizmo_rotation: CameraRotation, 
}
```

editor/src/scene_viewer/gizmo.rs
```diff
@@ -267,41 +275,56 @@ impl SceneGizmo {
...
+                        default_color
+                    },
+                );
+            }
         }
     }
 
-    pub fn on_click(&self, pos: Vector2<f32>, engine: &Engine) -> Option<SceneGizmoAction> {
+    pub fn on_click(&mut self, pos: Vector2<f32>, engine: &Engine) -> Option<SceneGizmoAction> {
+        self.first_click_pos = pos;
```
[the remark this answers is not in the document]
NathanaelG1: Yeah.. let me take that out.

editor/src/scene_viewer/gizmo.rs
```diff
@@ -267,41 +275,56 @@ impl SceneGizmo {
...
+        if self.is_dragging {
+            return Some(SceneGizmoAction::Rotate(CameraRotation {
+                pitch: self.first_pitch.to_radians(),
+                yaw: self.first_yaw.to_radians(),
```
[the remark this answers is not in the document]
NathanaelG1: Ah, I think this was occasionally causing some misalignment, seems very smooth now!

editor/src/scene_viewer/gizmo.rs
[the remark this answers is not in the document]
mrDIMAS: You can do this:

```rust
struct DragContext {
    pub initial_click_pos: Vector2<f32>,
    pub initial_rotation: CameraRotation, 
}
```

and in the gizmo:

```rust
drag_context: Option<DragContext>
```

So you don't need to have `is_dragging` flag, because you can do `if let Some(drag_context) = self.drag_context.as_ref() {..}`.

NathanaelG1:
Overall the functionality is looking pretty good. I wanted to get a clear guideline for how the feature should behave. For instance, in Blender (just as an example, not trying to make this like Blender 🦊   ) if you click on a spot in the rotation it will gently glide the camera to that orientation. 

https://github.com/user-attachments/assets/56deb976-4ba7-4335-8021-9f6c1363c99b


### On Click 
Right now doing that in Fyrox immediately shows that perspective similar to clicking on a gizmo cone. I have also been able to get it to not move at all on click, and only move with the mouse on click.  Do we want: 
  1. Snap to perspective
  2. Glide to perspective

### On Mouse Move
Should the mouse be moving a certain speed before rotating the gizmo? 

Should the action stop if the mouse leaves the gizmo window? Or should it only stop when the mouse button is released no matter where on the editor window? 

### Motion stabilizing
Things work smoothly and is very helpful to me to be able to rotate perspectives quickly. Should rotation be locked in any way based on the orientation of the gizmo? For instance if x axis is vertical clamp the yaw or if y axis is vertical clamp the pitch to 89 like you have in scene_gizmo.on_click()? 

### Key Bindings
Do we want to add any key bindings linked in settings so that its configurable?

NathanaelG1:
I have also noticed even in the release version, if you move the camera too much or click the gizmo too much it can occasionally get locked. I haven't looked into exactly what's happening when it locks but I usually have to perform some clicks in random places or click a lot on the gizmo until something makes it through. I can also look into this unless the locking is happening for a specific reason.

editor/src/scene_viewer/gizmo.rs
```diff
@@ -267,41 +274,53 @@ impl SceneGizmo {
...
+        &mut self,
+        pos: Vector2<f32>,
+        engine: &mut Engine,
+        camera_controller: &mut CameraController,
+    ) {
+        if let Some(drag_context) = self.drag_context.as_ref() {
+            let delta = pos - drag_context.initial_click_pos;
+            let sens: f32 = 0.09;
+            camera_controller.yaw = drag_context.initial_rotation.yaw + delta.x * -sens;
+            camera_controller.pitch = drag_context.initial_rotation.pitch + delta.y * sens;
```
mrDIMAS, line 290: I completely forgot - here you need to clamp the pitch in `-90.0..90.0` degree range. However, since all angles are in radians, the range will be `-90.0.to_radians()..90.0.to_radians()`. This is why you can rotate the camera upside down using scene gizmo.

editor/src/scene_viewer/gizmo.rs
```diff
@@ -267,41 +274,53 @@ impl SceneGizmo {
...
-                node,
-                graph,
-                if node == closest {
-                    Color::opaque(255, 255, 0)
-                } else {
-                    default_color
-                },
-            )
+    pub fn on_click(&mut self, pos: Vector2<f32>, engine: &Engine) -> Option<SceneGizmoAction> {
+        if let Some(_drag_context) = self.drag_context.as_ref() {
```
mrDIMAS, line 319: `if self.drag_context.is_some() {`

editor/src/scene_viewer/mod.rs
```diff
@@ -790,37 +788,54 @@ impl SceneViewer {
...
-                                                                self.camera_projection,
-                                                                MessageDirection::ToWidget,
-                                                                Some(0),
-                                                            ),
-                                                        );
-                                                    }
+                                    self.scene_gizmo.drag_context = Some(gizmo::DragContext {
+                                        initial_click_pos: rel_pos,
+                                        initial_rotation: gizmo::CameraRotation {
+                                            pitch: game_scene.camera_controller.pitch.to_radians(),
```
mrDIMAS, line 794: Why do you need to convert pitch and yaw here to radians? If you look closely at the camera controller code, you'll find that these angles are already in radians.

mrDIMAS (changes requested):

editor/src/scene_viewer/mod.rs
NathanaelG1: I did check out the camera controller some but when I was logging the positions the numbers reported were not in radians, since I adjusted the rotation messaging this may not be happening anymore. Ill test it out and let you know.

mrDIMAS:
Now onto your questions: 

> 2\. Glide to perspective

Camera gliding and all kinds of animations in UI are just distractions and takes user's time for nothing. It may look cool for a few times, but when you'll start using it hundreds of times per day you'll be annoyed that some animation does not let you do a thing instantly.

> Should the mouse be moving a certain speed before rotating the gizmo?

Probably no, because when you pressed a button and started dragging the mouse you're already confirmed (to the UI) that you're dragging and scene gizmo should rotate the camera.


> Should the action stop if the mouse leaves the gizmo window?

No, there's a concept of mouse capturing in UI - it basically says that when you pressed a button on a widget, then this widget will receive mouse events no matter if the mouse is actually above it or not.

editor/src/scene_viewer/gizmo.rs
NathanaelG1: Great! I can do that. Was also curious about leaving the sens variable there, my intuition was to set it as a const but I haven't noticed many consts in my perusal of the engine. I could also just place the number but I felt that labelling it gave it extra readability. 

 Eventually there could be finer grained controls (allowing the user to turn off dragging, etc etc) of the gizmo so the sensitivity might be good to place into the drag_context as a const.

NathanaelG1:
Sounds good and totally agree, ill get this cleaned up in the morning. Thank you.

editor/src/scene_viewer/mod.rs
NathanaelG1: Yep removing radians works doesn't affect functionality now that the messaging is set up properly.

PR #673, opened 2024-08-03
New terrain version with height texture margins and normals in shader

b-guild:
Previously, terrain rendering had no normal data, with normals always pointing directly up. In order to resolve this problem, I have increased the terrain's version number from 1 to 2, and in the new version all height textures have a one-pixel margin with data that is copied from the neighboring chunks. This allows each terrain vertex to calculate its normal value based on the nearby heights.

This required a significant increase in complexity for the terrain shader and the terrain node in order to correctly account for the margins. During this process I have encountered numerous problems and fixed each one, but due to the complexity of the the whole project it is difficult to be totally confident that there are no unknown issues still remaining. At best I can say that everything *seems* to finally be working correctly, and it is so important for terrain to have correct normals.

While I was updating the visit method for the terrain, I modified the version system so that the terrain node no longer stores its version number. All terrain nodes currently in memory should always have the most current version. The version number is now just stored in the visitor, because that is all that is required in order for future terrain nodes to do conversions to update older versions.

I created a `validate` method for terrain nodes to alert users to the very particular requirements for height textures sizes and block sizes. It is practically impossible to allow total freedom in choosing these sizes without issues, so we should be firm that height textures need to be a power of 2 plus 3 on each dimension, both to allow room for the margins and to allow the texture to be recursively split into quads. For similar reasons, blocks should be a power of 2 plus 1 in each dimension, and obviously blocks should not be larger than the height texture.

I removed the `update_height_pixel` method that I previously created since it was originally misguided, unnecessary, and making it work with the new new height textures would have added more complexity and potential for bugs in an already error-prone project.

### DIFF
--- a/fyrox-impl/src/material/shader/standard/terrain.shader
+++ b/fyrox-impl/src/material/shader/standard/terrain.shader
@@ -135,13 +135,24 @@
                     // Each node has tex coords in [0; 1] range, here we must scale and offset it
                     // to match the actual position.
                     vec2 actualTexCoords = vec2(vertexTexCoord * nodeUvOffsets.zw + nodeUvOffsets.xy);
-                    float height = texture(heightMapTexture, actualTexCoords).r;
+                    vec2 heightSize = vec2(textureSize(heightMapTexture, 0));
+                    vec2 innerSize = heightSize - 3.0;
+                    vec2 pixelSize = 1.0 / heightSize;
+                    vec2 heightCoords = (actualTexCoords * innerSize + 1.5) * pixelSize;
+                    float height = texture(heightMapTexture, heightCoords).r;
                     vec4 finalVertexPosition = vec4(vertexPosition.x, height, vertexPosition.z, 1.0);
+                    float hx0 = texture(heightMapTexture, heightCoords + ivec2(-1, 0) * pixelSize, 0).r;
+                    float hx1 = texture(heightMapTexture, heightCoords + ivec2(1, 0) * pixelSize, 0).r;
+                    float hy0 = texture(heightMapTexture, heightCoords + ivec2(0, -1) * pixelSize, 0).r;
+                    float hy1 = texture(heightMapTexture, heightCoords + ivec2(0, 1) * pixelSize, 0).r;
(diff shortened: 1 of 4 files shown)


fyrox-impl/src/material/shader/standard/terrain.shader
```diff
@@ -206,8 +217,8 @@
 
                     outColor = diffuseColor * texture(diffuseTexture, tc);
 
-                    vec4 n = normalize(texture(normalTexture, tc) * 2.0 - 1.0);
-                    outNormal = vec4(normalize(tangentSpace * n.xyz) * 0.5 + 0.5, 1.0);
```
mrDIMAS, line 210: Do not remove `* 0.5 + 0.5` - it is needed because normals in the GBuffer are stored in RGBA8 texture, which has 0..255 value range. When you do `* 0.5 + 0.5` you shifting `-1.0..1.0` range to `0.0..1.0` and then it will be unpacked back when reading from the GBuffer.

fyrox-impl/src/scene/terrain/mod.rs
```diff
@@ -1118,13 +1271,188 @@ fn pixel_position_to_grid_position(
...
+}
+
 impl TypeUuidProvider for Terrain {
     fn type_uuid() -> Uuid {
         uuid!("4b0a7927-bcd8-41a3-949a-dd10fba8e16a")
     }
 }
 
 impl Terrain {
+    /// The height map of a chunk must have one-pixel margins around the edges which do not correpond
```
mrDIMAS, line 1335: Typo `correspond`

mrDIMAS (changes requested):

fyrox-impl/src/material/shader/standard/terrain.shader
b-guild: I have done some experiments to see the effect of removing the * 0.5 + 0.5. At first I found nothing apparent, but by greatly increasing the intensity of the light and flattening the terrain, the problem became dramatically visible. It is disturbing that I didn't not see the problem immediately. Now I am dubious of every shadow, questioning whether it is as it should be or whether it is a glitch. I hope people better at shaders than I continue to work on this.

Issue #351, opened 2022-08-22
Add an ability to cut holes from terrain geometry


mrDIMAS:
There is no way to cut holes in terrains geometry, so there is no way create cave entrances in the terrain. This may be a real blocker for some games. 

It is also important to add proper tooling for this. The editor should have special brush, that allows to cut holes of various shapes.

b-guild:
Is the goal to make part of the material transparent, or is the goal to actually remove parts of the mesh?

There are costs associated with rendering transparent materials, but aside from that making part of the terrain transparent ought to be technically less difficult.

Removing part of the mesh means the terrain does not need to be rendered as transparent, but it also means we have to store more mesh data as it cannot simply use repeated instances of that one mesh block. There would be limits to the available hole shapes and we would need a data structure to store the holes. We would have to think about how the holes interact with the LOD system. Are the holes always rendered at max LOD? Are there different versions of the holes for different levels of detail?

mrDIMAS:
I think it should be purely visual, which means that there should be a mask for holes. Also, such holes should be taken into account when making height field collider for terrains.

b-guild:
Does this mean that the terrain will need to be rendered using forward rendering in order to make the holes transparent?

mrDIMAS:
No, check terrain default material, it is already rendering with alpha blending into G-Buffer. This is needed to support multiple layers. Holes will act like a simple alpha test and that should be enough.

b-guild:
If it is that easy, then why does the documentation say that forward rendering is necessary for transparency? The rendering section of the Fyrox book says that forward rendering "should be only used to render translucent objects." The documentation for scene::mesh::RenderPath::Deferred says "Deferred rendering has much better performance than Forward, but it does not support transparent objects and there is no way to change blending."

The documentation for Unity also claims that their deferred shading render path cannot handle transparency and that forward shading is necessary in that case.

I am no expert at shaders, but the GBuffer pass of the terrain shader has depth_write set to true, and that seems like it could cause trouble for transparency, as it should stop anything further away from being rendered regardless of alpha.

It should be possible to prevent pixels from being written using the `discard` GLSL command, but it seems that including that command anywhere in a fragment shader seriously reduces performance, and so it is recommended to be used only on small meshes.

mrDIMAS:
[...]

PR #676, opened 2024-08-15
Terrain holes

b-guild:
Introducing holes to the terrain. This involves:

* A new command: `ModifyTerrainHolesCommand` which is almost exactly like `ModifyTerrainHeightCommand`.
* A new brush target `HoleMask`.
* Modifications to the standard terrain shader to allow it to accept the hole mask and to discard pixels within the holes.
* Modifications to how terrain colliders are generated, including ensuring that the collider is aligned with the terrain so long as the collider node is at the same position as the terrain node.
* Modifications to how terrain textures are resizes so that the hole mask texture size is kept synchronized with the height map texture size.

Nothing has been done to help the collider stay in sync with changes to the terrain. That is can be work for a future commit.

### DIFF
--- a/editor/src/interaction/terrain.rs
+++ b/editor/src/interaction/terrain.rs
@@ -68,7 +68,9 @@ use crate::{
     make_color_material,
     message::MessageSender,
     scene::{
-        commands::terrain::{ModifyTerrainHeightCommand, ModifyTerrainLayerMaskCommand},
+        commands::terrain::{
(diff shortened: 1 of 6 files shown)


mrDIMAS (approved):

Vlad2001MFS:
[#351](https://github.com/FyroxEngine/Fyrox/issues/351) related

PR #683, opened 2024-10-09
Refactor hot reload to allow custom dynamic plugins besides dylib-based

kkolyan:
`AbstractDynamicPlugin` trait introduced as a contract between engine and dynamic plugins. All dylib-specific code is hidden inside `DyLibPlugin`, the only built-in implementation of the `AbstractDynamicPlugin`.

### DIFF
--- a/fyrox-impl/src/engine/mod.rs
+++ b/fyrox-impl/src/engine/mod.rs
@@ -42,8 +42,6 @@ use crate::{
         futures::{executor::block_on, future::join_all},
         instant,
         log::Log,
-        notify,
-        notify::{EventKind, RecursiveMode, Watcher},
         pool::Handle,
         reflect::Reflect,
         task::TaskPool,
@@ -65,7 +63,7 @@ use crate::{
         Material,
     },
     plugin::{
-        dynamic::DynamicPlugin, DynamicPluginState, Plugin, PluginContainer, PluginContext,
+        AbstractDynamicPlugin, Plugin, PluginContainer, PluginContext,
         PluginRegistrationContext,
     },
     renderer::{framework::error::FrameworkError, Renderer},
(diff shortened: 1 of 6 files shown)


fyrox-impl/src/engine/mod.rs
```diff
@@ -2413,25 +2296,22 @@ impl Engine {
...
+        let PluginContainer::Dynamic(plugin) = plugin_container
         else {
             return Err(format!(
                 "Plugin {plugin_index} is static and cannot be reloaded!",
             ));
         };
 
-        if matches!(state, DynamicPluginState::Unloaded { .. }) {
+        if !plugin.is_loaded() {
+            // TODO: this means that something bad happened during plugin reloading.
```
mrDIMAS, line 2307: Yeah, clearing `need_reload` flag is the way to go.

mrDIMAS:
Looks good, please test your changes thoroughly, these dynamic plugins sucked a lot of blood from me and I don't want to debug them yet again.

mrDIMAS:
I'm ok with that, feel free to rename it if you think new names are better.

mrDIMAS:
Could you please fix CI errors? It seems that there's one warning treated like an error and it should be easy to fix.

fyrox-impl/src/plugin/dylib.rs
```diff
@@ -0,0 +1,334 @@
...
+
+    /// Return a reference to the plugin interface of the dynamic plugin.
+    pub(crate) fn plugin_mut(&mut self) -> &mut dyn Plugin {
+        &mut *self.plugin
+    }
+}
+
+
+/// Implementation of DynamicPluginTrait that [re]loads Rust code from Rust dylib .
+pub struct DyLybDynamicPlugin {
```
mrDIMAS, line 98: Typo `DyLyb...`, should be `DyLib...`

mrDIMAS (changes requested):

Issue #696, opened 2024-10-29
Scene settings reset to default on scene save through the editor


guydunigo:
Good evening,

# Problem

When I change the scene settings in the editor (disable physics 2D, change gravity, ...) and then save the file. If I open those settings again, the modified values are still there, but they are reset to default if I close the file and open it again. So those modifications are not saved to the file.
If I force change the binary scene file (e.g. hexedit) and load it in the editor, the modified values are correctly displayed, but saving the file (even with no modifications) puts scene settings back to default.

# Version

This happens on 34.1 and on Fyrox nightly commit 77e0a89a.

# Analysis

The problem seems to come from `make_purified_scene` (`editor/src/scene/mod.rs`) which calls `Graph::clone` (`fyrox-impl/src/scene/graph/mod.rs`),
but this method doesn't clone the PhysicsWorld (2D and 3D).
Sadly PhysicsWorld doesn't implement clone and I don't know your code enough nor how much *purification* was intended to directly create a PR for you.

There might also be other non-cloned properties that should be saved as well.

# Temporary workaround

Doing it programmatically in `on_scene_loaded`.

mrDIMAS:
Looks like it was simply forgotten, some parameters from both physics worlds must be copied indeed.

PR #697, opened 2024-10-29
Cloning physics when cloning Graph to persist Scene settings when sav…

guydunigo:
…ing Scene from the editor

Trying to fix : https://github.com/FyroxEngine/Fyrox/issues/696

This clones properties that are not `#[visit(skip)]` because it's the same mechanism as far as I understand it.

As mentioned in the issue, this might not fit the engine design and there might be other properties in the scene that were forgotten to clone.

*By the way, since Visiting is a form of cloning, isn't there a way to implement automatically Clone based on Visit impl (via macro or extending trait) ?*

### DIFF
--- a/fyrox-impl/src/scene/dim2/physics.rs
+++ b/fyrox-impl/src/scene/dim2/physics.rs
@@ -506,6 +506,17 @@ pub struct PhysicsWorld {
     debug_render_pipeline: Mutex<DebugRenderPipeline>,
 }
 
+impl Clone for PhysicsWorld {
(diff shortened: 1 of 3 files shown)


mrDIMAS:
Looks good, I can merge it right away. 

> By the way, since Visiting is a form of cloning, isn't there a way to implement automatically Clone based on Visit impl (via macro or extending trait) ?

Visiting is not cloning, it is serialization/deserialization. It is just a bad naming.

PR #698, opened 2024-11-05
Fix for UI layout, including Grid and Text.

b-guild:
I have implemented the WPF algorithm for laying out Grid widgets, including the resolution system for cases that involve some widgets being measured twice in order to determine some column widths.

I have expanded on the documentation for Grid and made the meanings of Strict, Auto, and Stretch more precise. A few grids in the editor were clearly using the wrong dimensions for their rows/columns, and I have made the necessary corrections so that the editor UI is not affected by these changes to Grid.

Part of these fixes to the editor's grid dimensions was fixing the dimensions of AssetItem's grid, and in order to make AssetItem work as it should I had to fix a problem with Text that was inserting some extra blank lines. This no longer happens, and I have added some unit tests that would have caught this bug. Now AssetItems can have letter wrap for their text widgets, allowing longer file names to be shown.

### DIFF
--- a/editor/src/asset/item.rs
+++ b/editor/src/asset/item.rs
@@ -326,13 +326,14 @@ impl AssetItemBuilder {
                                         .with_margin(Thickness::uniform(1.0))
                                         .on_row(1),
                                 )
+                                .with_wrap(WrapMode::Letter)
                                 .with_horizontal_text_alignment(HorizontalAlignment::Center)
                                 .with_text(path.file_name().unwrap_or_default().to_string_lossy())
                                 .build(ctx),
                             ),
                     )
-                    .add_column(Column::auto())
-                    .add_row(Row::stretch())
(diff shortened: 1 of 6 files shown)


mrDIMAS:
Cool, I need some time to review it.

mrDIMAS (changes requested):

mrDIMAS:
I checked the changes on my PC and found that World Viewer's items aren't filling all the space available for them: 
изображение

b-guild:
That should take care of it. It was easier than I thought. The changes to Grid do slightly modify its behavior, but they make it work the way it should. When something is supposed to fill the available space, it should be Stretch. When something is supposed to be at its natural size, it should be Auto. Often Auto seems to be used when it should be Stretch, and that is the source of these issues. In this case the Auto column was in Tree, and I changed it to Stretch.

PR #724, opened 2025-02-06
Add additional mass properties

ttwoti:
This PR adds `mass_properties_type` variable of `RigidBodyMassPropertiesType` type to the RigidBody and makes it available to edit `center_of_mass` and `principal_inertia` properties in the editor.

Allows to change those properties when the default values are not great, for example when it is required to have center of mass at a known predefined position.

### DIFF
--- a/fyrox-impl/src/scene/rigidbody.rs
+++ b/fyrox-impl/src/scene/rigidbody.rs
@@ -146,6 +146,21 @@ pub(crate) enum ApplyAction {
     WakeUp,
 }
 
+#[derive(Copy, Clone, Debug, Reflect, Visit, PartialEq, AsRefStr, EnumString, VariantNames)]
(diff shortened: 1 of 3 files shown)


fyrox-impl/src/scene/rigidbody.rs
```diff
@@ -200,6 +215,9 @@ pub struct RigidBody {
     #[reflect(setter = "set_gravity_scale")]
     pub(crate) gravity_scale: InheritableVariable<f32>,
 
+    #[reflect(setter = "set_mass_properties_type")]
```
mrDIMAS, line 218: Add `#[visit(optional)]` attribute to keep backward compatibility.

ttwoti: Added in 4fbaa6d

PR #733, opened 2025-02-16
Add support for additional audio formats

### LINKED ISSUE #707: Migrate to Symphonia to support more audio formats
A very common format and it should be supported by the sound system.

**Edit.** Symphonia looks to be a better replacement for current decoder implementation.

septante:
Resolves #707

### DIFF
--- a/fyrox-sound/src/buffer/mod.rs
+++ b/fyrox-sound/src/buffer/mod.rs
@@ -47,6 +47,7 @@ use std::{
     path::{Path, PathBuf},
     time::Duration,
 };
+use symphonia::core::io::MediaSource;
 
 pub mod generic;
 pub mod loader;
@@ -178,6 +179,16 @@ impl Seek for DataSource {
                 unreachable!("Raw data source does not supports Seek trait!")
             }
         }
+    }
+}
+
+impl MediaSource for DataSource {
+    fn is_seekable(&self) -> bool {
+        true
+    }
+
+    fn byte_len(&self) -> Option<u64> {
+        None
     }
 }
 
--- a/fyrox-resource/src/io.rs
+++ b/fyrox-resource/src/io.rs
@@ -32,9 +32,9 @@ use std::{
 };
 
 /// Trait for files readers ensuring they implement the required traits
-pub trait FileReader: Debug + Send + Read + Seek + 'static {}
+pub trait FileReader: Debug + Send + Sync + Read + Seek + 'static {}
 
-impl<F> FileReader for F where F: Debug + Send + Read + Seek + 'static {}
+impl<F> FileReader for F where F: Debug + Send + Sync + Read + Seek + 'static {}
 
 /// Interface wrapping IO operations for doing this like loading files
 /// for resources
--- a/fyrox-sound/Cargo.toml
+++ b/fyrox-sound/Cargo.toml
@@ -17,11 +17,10 @@ rust-version = "1.80"
 [dependencies]
 fyrox-core = { path = "../fyrox-core", version = "0.36.0" }
 fyrox-resource = { path = "../fyrox-resource", version = "0.36.0" }
-lewton = "0.10.2"
-ogg = "0.8.0"
 hrtf = "0.8.0"
 hound = "3.4.0"
 strum = "0.26.1"
 strum_macros = "0.26.1"
 tinyaudio = "1"
 serde = { version = "1", features = ["derive"] }
+symphonia = "0.5.4"
--- a/fyrox-sound/src/decoder/vorbis.rs
+++ b/fyrox-sound/src/decoder/vorbis.rs
@@ -72,42 +72,21 @@ impl Iterator for OggDecoder {
     }
 }
 
-fn is_vorbis_ogg(mut source: DataSource) -> bool {
-    let pos = source.stream_position().unwrap();
+impl OggDecoder {
+    // TODO: fix return type
+    pub fn new(mut source: DataSource) -> Result<Self, ()> {
+        let channel_duration_in_samples;
 
-    let media_source =
-        MediaSourceStream::new(Box::new(source), MediaSourceStreamOptions::default());
+        let initial_stream_position = source.stream_position().unwrap();
 
-    let res = OggReader::try_new(media_source, &FormatOptions::default());
+        let media_source =
+            MediaSourceStream::new(Box::new(source), MediaSourceStreamOptions::default());
+        let mut reader = OggReader::try_new(media_source, &FormatOptions::default()).unwrap();
+        let mut decoder =
+            VorbisDecoder::try_new(&CodecParameters::default(), &DecoderOptions::default())
+                .unwrap();
 
-    if let Ok(mut reader) = res {
-        reader
-            .seek(
-                SeekMode::Accurate,
-                SeekTo::Time {
-                    time: Time::new(pos, 0.0),
-                    track_id: None,
-                },
-            )
-            .unwrap();
-
-        true
-    } else {
-        false
-    }
-}
-
-// God bless `stb_vorbis` - https://github.com/nothings/stb/blob/master/stb_vorbis.c#L4946
-// lewton::audio::get_decoded_sample_count is bugged and does not work correctly. So instead of using it,
-// we use `stb_vorbis` approach - find last packet, take its position and return it. This function is still
-// unideal, because we read all packets one-by-one, instead of just jumping to the last one.
-fn total_duration_in_samples(mut source: DataSource) -> usize {
-    let initial_stream_position = source.stream_position().unwrap();
-
-    let media_source =
-        MediaSourceStream::new(Box::new(source), MediaSourceStreamOptions::default());
-
-    if let Ok(mut reader) = OggReader::try_new(media_source, &FormatOptions::default()) {
+        // Get duration
         let mut last_packet = None;
         while let Ok(packet) = reader.next_packet() {
             last_packet = Some(packet);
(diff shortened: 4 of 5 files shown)


mrDIMAS:
IMO the all submodules of `decoder` module should be deleted and internals of `impl Decoder` should be replaced with `symphonia`. It does not really worth it to replace decoders one by one.

septante:
Added a generic implementation and removed the submodules. Still need to add an error type, but it seems like the tests pass and the play_sound example mentioned in the issue still runs.

mrDIMAS:
I think it can be done in this PR.

septante:
Should I change the callers to return Errors as well, or try to keep the original signatures that return DataSource as the Err type?

septante:
Changed the signatures to Errors for now and fixed all the warnings/errors. The dependency on hound for wav files can't be removed for now, since it is used in the write_wav example. Not sure if the example should be rewritten or just left as is.

fyrox-sound/src/decoder/mod.rs
```diff
@@ -18,84 +18,172 @@
...
+        let res = probe.format(
+            &Hint::new(),
+            media_source_stream,
+            &FormatOptions::default(),
+            &MetadataOptions::default(),
+        )?;
+
+        let mut reader = res.format;
+        let tracks = reader.tracks();
+        let first_track = tracks.first().unwrap();
```
mrDIMAS, line 97: Replace `unwrap` call with error mapping (`ok_or_else`) and use `?` with it.

mrDIMAS (changes requested):

fyrox-sound/src/decoder/mod.rs
```diff
@@ -18,84 +18,172 @@
...
+        let mut vec: Vec<f32> = Vec::new();
+        if let Ok(packet) = reader.next_packet() {
+            if let Ok(decoded) = decoder.decode(&packet) {
+                let buffer: AudioBuffer<f32> = decoded.make_equivalent();
+                let samples = buffer.chan(0);
+
+                vec = samples.to_vec();
+            }
+        }
+        let samples = vec.into_iter();
```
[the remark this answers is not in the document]
septante: Done

fyrox-sound/src/decoder/mod.rs
```diff
@@ -18,84 +18,172 @@
...
+            .seek(
+                SeekMode::Accurate,
+                SeekTo::Time {
+                    time: location.into(),
+                    track_id: None,
+                },
+            )
+            .is_err()
+        {
+            println!("Failed to seek vorbis/ogg?")
```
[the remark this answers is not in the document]
septante: Should this function return the error instead? It seems a little odd that rewind does but seek doesn't.

fyrox-sound/src/decoder/mod.rs
```diff
@@ -18,84 +18,172 @@
...
-        match self {
-            Decoder::Wav(wav) => wav.next(),
-            Decoder::Ogg(ogg) => ogg.next(),
+        if let Some(sample) = self.samples.next() {
+            Some(sample)
+        } else {
+            if let Ok(packet) = self.reader.next_packet() {
+                if let Ok(decoded) = self.decoder.decode(&packet) {
+                    let buffer: AudioBuffer<Self::Item> = decoded.make_equivalent();
+                    let samples = buffer.chan(0);
```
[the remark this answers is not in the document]
septante: It seems like
```rust
let left_samples = buffer
    .chan(Channels::FRONT_LEFT.bits() as usize)
    .into_iter();
let right_samples = buffer
    .chan(Channels::FRONT_RIGHT.bits() as usize)
    .into_iter();
```
and then interleaving should work? Either with the itertools crate or a hand-written implementation.

fyrox-sound/src/decoder/mod.rs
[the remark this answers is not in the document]
mrDIMAS: Yeah, you're right.

fyrox-sound/src/decoder/mod.rs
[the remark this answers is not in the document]
septante: Would iterating over the channels and interleaving all of the slices be a reasonable approach?

fyrox-sound/src/decoder/mod.rs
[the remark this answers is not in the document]
septante: Done

fyrox-sound/src/decoder/mod.rs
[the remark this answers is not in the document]
mrDIMAS: It may work, but it probably will be relatively slow to read than from an interleaved array.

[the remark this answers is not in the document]
septante: Implemented stereo audio, though when trying to play the file found [here](https://www.aoakley.com/articles/2018-08-24-stereo-test.php) I noticed that it only plays the first "front center", when the entire clip includes side left/right, followed by repeating each of the three one more time. Not sure if this is a limitation of symphonia or if there's something else I'm missing.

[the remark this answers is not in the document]
septante: Never mind, the thread wasn't sleeping for long enough. I think that's everything done for this PR?

fyrox-sound/src/decoder/mod.rs
```diff
@@ -131,11 +131,30 @@ impl Decoder {
...
         let mut buffer: AudioBuffer<f32> = decoded.make_equivalent();
         decoded.convert(&mut buffer);
-        let samples = buffer.chan(0).to_vec();
+        let samples = Self::interleaved(&mut buffer);
 
         Ok(samples.into_iter())
     }
 
+    fn interleaved(buffer: &mut AudioBuffer<f32>) -> Vec<f32> {
+        let channels = buffer.spec().channels.count();
```
mrDIMAS, line 151: Naming is confusing here. This should be `channel_count`.

fyrox-sound/src/decoder/mod.rs
```diff
@@ -131,11 +131,30 @@ impl Decoder {
...
+        let frames = buffer.frames();
+
+        let mut samples = Vec::with_capacity(channels);
```
mrDIMAS, line 154: This should be `channels`

mrDIMAS (changes requested):

fyrox-sound/src/decoder/mod.rs
septante: Done

fyrox-sound/src/decoder/mod.rs
septante: Done

mrDIMAS:
Could you please check if all examples from `fyrox-sound` are working (especially the streaming one)? If everything is ok, I'll merge the PR right away.

septante:
Hm, streaming appears to be broken. I'll take a look

septante:
Some notes so I don't forget: the cause of the panic seems to be an unwrap occurring [here](https://docs.rs/symphonia-format-ogg/0.5.4/src/symphonia_format_ogg/demuxer.rs.html#124), because the check [here](https://docs.rs/symphonia-format-ogg/0.5.4/src/symphonia_format_ogg/demuxer.rs.html#308) returns None. The solution should just be to change the byte_len implementation in buffer/mod.rs.

septante:
Fixed! I wasn't sure if I could use the length of the vector as the byte_len value for DataSouce::Memory, so I just used None for now. I left a TODO note in the last commit mentioning it, if you want to take a look.

fyrox-sound/src/buffer/mod.rs
```diff
@@ -181,15 +187,63 @@ impl Seek for DataSource {
...
+        match self {
+            DataSource::File { .. } | DataSource::Memory(_) => true,
+            DataSource::Raw { .. } | DataSource::RawStreaming(_) => false,
+        }
+    }
+
+    fn byte_len(&self) -> Option<u64> {
+        match self {
+            DataSource::File { path, data: _ } => {
+                if let Ok(f) = File::open(path) {
```
mrDIMAS, line 201: Don't use `File::open` here, it is not compatible with neither Android nor Wasm. Instead, use `MediaSource::byte_len(data)` itself, it should work since data is `Box<dyn FileReader>` which implements all the traits for `MediaSource` trait.

fyrox-sound/src/buffer/mod.rs
```diff
@@ -181,15 +187,63 @@ impl Seek for DataSource {
...
+                    MediaSource::byte_len(&f)
+                } else {
+                    None
+                }
+            }
+            // TODO: can we use the length of the underlying vector here?
```
mrDIMAS, line 207: Yeah, you can use the length here.

mrDIMAS (changes requested):

fyrox-sound/src/buffer/mod.rs
septante: I'm not seeing an implementation for MediaSource on FileReader? Trying to use MediaSource::byte_len(data) fails to compile.

fyrox-sound/src/buffer/mod.rs
septante: Fixed

fyrox-sound/src/buffer/mod.rs
mrDIMAS: You can always do this 

```rust
let byte_len = data.seek(SeekFrom::End(0))?;
data.seek(SeekFrom::Start(0))?;
```

septante: Hm, that doesn't seem to work either, since self is not a mut reference.

mrDIMAS: Ok, just add a `byte_len` method to `FileReader` and implement the trait for `std::fs::File` and remove auto-impl for it. The impl could be this https://docs.rs/symphonia-core/latest/src/symphonia_core/io/mod.rs.html#69

septante: Should I add a default implementation that returns None to avoid breaking other uses of FileReader?

septante: Oh, it still needs to be implemented for the other types regardless.

mrDIMAS (approved):

PR #736, opened 2025-02-21
Introducing tile map brush macros and autotile

b-guild:
I have created a system for adding automation to tile map editing that I have called "macros," and to make sure that the system is effective and practical I have created two macros that use the system in different ways.

* Autotiling allows the user to paint with terrains instead of tiles. Each terrain represents a set of various tiles with patterns to tell the macro which tiles are allowed to be adjacent to which other tiles. The macro automatically switches to the appropriate tiles as you draw, saving you from having to manually find the tile you want. I should also be possible to make this work while the game is running, but I have not tried that.
* Wave Function Collapse allows the user to paint a region of the tile map, and then the macro automatically fills the region with randomly generated tiles using patterns to ensure that the randomly selected tiles always fit together. This should allow users to quickly fill areas with interesting looking tiles without needing to be creative. The random generation sometimes fails if the algorithm paints itself into a corner and cannot find a tile that will fit into some cell, so it can be configured to automatically retry a certain number of times and hopefully one of those attempts will randomly succeed. Future improvements may allow it to start a task that works on this in the background instead of blocking the editor for a moment while it works.

These macros seem to work excellently 95% of the time, but I have noticed that the autotile macro sometimes goes wrong and I have been unable to understand what causes this or how to trigger it. Hopefully a solution will be found shortly.

Macros are accessible using the tile map brush editor which now has a new "Macros" tab which reveals the macro editor which allows macro instances to be added, removed, and their settings changed. To activate a macro you must select a cell in the brush and add it to the macro to indicate that painting with this brush cell should activate the macro instead of doing normal tile map painting. This can be done in the inspector area of the tile map brush editor.


editor/src/plugins/tilemap/brush_macro.rs
```diff
@@ -0,0 +1,992 @@
...
+                    TileSetPropertyValueElement::F32(*v),
+                ));
+                send_sync_message(
+                    ui,
+                    DropdownListMessage::selection(
+                        self.list,
+                        MessageDirection::ToWidget,
+                        Some(find_list_index(prop, NamableValue::F32(*v))),
+                    ),
+                );
```
mrDIMAS, line 824: These three blocks looks almost identical and their difference is `TileSetPropertyValueElement` and `NamableValue` only, I think there should be a function that accepts such args and replace these three blocks with this function call.

mrDIMAS (changes requested):
Finally finished the review, there are only few spots that needs attention. In general it is hard to give a proper in-depth review with such amount of code, so I haven't really touched any algorithms and just checked the code for obvious issues.

mrDIMAS:
@b-guild Could you please add an explanation how these features work to the book? Maybe in a separate chapter, add some gifs and pics as well.

PR #749, opened 2025-03-11
Moving control panels to Inspector

b-guild:
This is an experiment to see if the editor is improved by moving some of the little control panel windows into the top of the Inspector. This has several benefits:

* It simplifies the code for this control panels, since they no longer need to manage their windows. This PR reduces the lines of code overall.
* It streamlines the Editor interface for the users, since they no longer need to manage these little windows.
* It resembles the UI of Unity, so it should be familiar to some users.

I considered doing the same with the control panels for Terrain and Tile Map, but I am less convinced this would be a good idea since these windows are larger. Instead I suggest that future work may allow a docking tile to hold multiple windows with tabs to select the currently visible window. This way these larger control panels can be docked in the same tile as the inspector and the user can tab to the control panel they want to use.

### DIFF
--- a/editor/src/audio/preview.rs
+++ b/editor/src/audio/preview.rs
@@ -18,8 +18,9 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+use fyrox::gui::widget::WidgetMessage;
+
 use crate::fyrox::graph::SceneGraph;
-use crate::fyrox::gui::HorizontalAlignment;
 use crate::fyrox::{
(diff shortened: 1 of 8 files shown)


mrDIMAS (changes requested):
Could you please rename `window` fields and variables to something else? This does not really seems to be correct. In general I like this approach, I checked how it looks and it indeed removes visual clutter.

b-guild:
What would be the best name? I agree that it's not a window anymore, but what is it?

b-guild:
What would be the best name? I agree that it's not a window anymore, but what is it?

mrDIMAS:
It could be a simple `control`, `root_widget` or similar.

Issue #729, opened 2025-02-11
Fix "teleportation" bug


mrDIMAS:
"Teleportation" is a bug when an object is drawn at (0,0,0) for one frame and on the one draws where it should be. It usually occurs when an object was just added to a scene.

PR #770, opened 2025-04-07
Adding NewAge to Node to solve teleport glitch

b-guild:
In regard to [the "teleportation" bug](https://github.com/FyroxEngine/Fyrox/issues/729#issue-2846142773), I think the underlying reason for the problem is a fundamental need for new nodes to not be rendered until the game has had a chance to prepare them, such as by moving them into their desired places and sending messages to them and so on. This means we should delay fully enabling them until *after* the node's scripts have had a chance to run their methods to do their first update and their first message handling.

This is not just about teleportation. It is about spending a frame to make sure that objects settle down into their desired shapes before they become an active part of the game world.

To this end, I have created a `NewAge` enum that keeps track of how long a node has been in the graph. It starts at `NewAge::FirstFrame`, which indicates a node that has just been added. At the end of the frame it advances to `NewAge::SecondFrame` which indicates that a new frame has started with this node already in it. This second frame gives the node a chance to run all its scripts and handle its messages. At the end of the second frame it advances to `NewAge::Old` which means the node is fully enabled so it renders and becomes part of the physics world.

This is *highly experimental.* It seems to work well in my tests, but I recommend that this should be tested in existing games to make sure it does not break anything.

### DIFF
--- a/fyrox-impl/src/scene/base.rs
+++ b/fyrox-impl/src/scene/base.rs
@@ -417,6 +417,38 @@ impl<T> DerefMut for TrackedProperty<T> {
     }
 }
 
+/// A measure of the the first frames of a node's life after it was added
+/// to the graph. Some nodes may not be ready to be seen by the player on
+/// the frame they are added, so we keep track of the age of new nodes to
+/// disable some features for one frame to give those nodes a chance to update
+/// themselves and process messages before they are fully enabled.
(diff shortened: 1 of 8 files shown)


b-guild:
I think I discovered a problem. This change may be interfering with the editor. Since nodes are not updated in the editor, added nodes remain new and are not rendered. I will have to investigate what can be done about this.

PR #782, opened 2025-04-18
Using a `BitMask` to control which nodes are visible to each camera

b-guild:
By giving each node a `BitMask` called `render_mask` we can provide fine control over whether a camera renders a particular node by doing a bitwise-and on the `render_mask` of the node and the camera. If the two masks have any set bits in common, then the camera will render the node as normal, but if the bitwise-and results in zero, then the node is treated as invisible.

`BitMask` is part of the collision system, since it is defined in `collider.rs`, but it can be useful for purposes like this beyond just collision. I have added some helper methods to make `BitMask` easier to use.

I have only tested this in 2D since the game that I am working on is a 2D game, but whether a game is 2D or 3D should make no difference to how `render_mask` works. So long as `render_mask` has all bits set, which is the default, this change should have no effect at all upon existing games.

### DIFF
--- a/fyrox-impl/src/scene/dim2/rectangle.rs
+++ b/fyrox-impl/src/scene/dim2/rectangle.rs
@@ -23,7 +23,6 @@
 //!
 //! See [`Rectangle`] docs for more info.
 
-use crate::scene::collider::BitMask;
 use crate::scene::node::constructor::NodeConstructor;
 use crate::{
     core::{
@@ -307,11 +306,7 @@ impl NodeTrait for Rectangle {
     }
 
     fn collect_render_data(&self, ctx: &mut RenderContext) -> RdcControlFlow {
-        if *self.render_mask & ctx.render_mask == BitMask::none() {
-            return RdcControlFlow::Continue;
-        }
-
-        if !self.should_be_rendered(ctx.frustum) {
+        if !self.should_be_rendered(ctx.frustum, ctx.render_mask) {
             return RdcControlFlow::Continue;
         }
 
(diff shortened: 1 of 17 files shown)


fyrox-impl/src/scene/dim2/rectangle.rs
```diff
@@ -306,6 +307,10 @@ impl NodeTrait for Rectangle {
     }
 
     fn collect_render_data(&self, ctx: &mut RenderContext) -> RdcControlFlow {
+        if *self.render_mask & ctx.render_mask == BitMask::none() {
+            return RdcControlFlow::Continue;
+        }
+
```
mrDIMAS, line 313: This should go into `should_be_rendered` method, because every scene node has a render mask.

mrDIMAS (changes requested):

fyrox-impl/src/scene/dim2/rectangle.rs
b-guild: I considered that, but it would require changing the parameters of `should_be_rendered` and I thought it was best to avoid modifying `NodeTrait` to minimize the potential impact of the change. Still, I see your point.

b-guild: I considered that, but it would require changing the parameters of `should_be_rendered` and I thought it was best to avoid modifying `NodeTrait` to minimize the potential impact of the change. Still, I see your point.

mrDIMAS (approved):

PR #783, opened 2025-04-22
Handle UI scaling in Screen Widget

TitanNano:
The only way to do UI scaling that I found was the way the editor is doing it, by setting the render transform on the UI root grid. This method also requires updating the grid width and height every time the window size changes.

Using the `Screen` widget is really convenient, but it is incompatible with the previously mentioned method of UI scaling. The UI always ends up double the window size and going out of bounds of the window. 

This made me decide to implement the UI scaling inside the `Screen` widget, which also seemed like an appropriate place to handle this. Automatic scaling can be turned on and off with a checkbox.

https://github.com/user-attachments/assets/cacac42f-6386-4f3e-986a-78861b4bded9

----

The implementation is not quite perfect. I would have preferred to update the layout transform inside the `update` function, but this function is not being called inside the editor, so I had to send a `UiMessage` in the `measure_override`

### DIFF
--- a/fyrox-ui/src/screen.rs
+++ b/fyrox-ui/src/screen.rs
@@ -33,6 +33,7 @@ use crate::{
     BuildContext, Control, UiNode, UserInterface,
 };
 
+use fyrox_core::algebra::Matrix3;
 use fyrox_graph::constructor::{ConstructorProvider, GraphNodeConstructor};
 use std::{
     cell::Cell,
@@ -114,10 +115,19 @@ use std::{
 pub struct Screen {
     /// Base widget of the screen.
     pub widget: Widget,
+
+    /// Will auto scale the UI size based on the current OS window scale.
+    pub auto_scale_ui: bool,
+
     /// Last screen size.
     #[visit(skip)]
     #[reflect(hidden)]
     pub last_screen_size: Cell<Vector2<f32>>,
+
+    /// Last UI scale factor.
+    #[visit(skip)]
+    #[reflect(hidden)]
+    pub last_visual_transform: Matrix3<f32>,
 }
 
 impl ConstructorProvider<UiNode, UserInterface> for Screen {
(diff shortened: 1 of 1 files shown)


mrDIMAS:
I don't really like the fact that the scaling factor is just stored inside the UserInterface instance without doing anything useful. A user of the library would expect that calling `set_scaling_factor` is actually doing something, but the current implementation does not scale anything.

mrDIMAS:
> The UI always ends up double the window size and going out of bounds of the window.

Is this only in the editor or in the game as well?

TitanNano:
> Is this only in the editor or in the game as well?

Only in the game. In the editor, the UI was never being scaled until now. 

The issue comes from the fact that `Screen` uses the physical number of pixels to set its size but the `layout_transform` will then scale the widget to screen_size * 2 (or any other scale factor)

mrDIMAS:
I think you should try to discard the scaling used in the `Screen` widget when doing `measure_override`. You should do:

```rust
let logical_size = transform_size(
	ui.screen_size(),
	&self.visual_transform().try_inverse().unwrap_or_default(),
);
```

and apply the scaling to the root UI node as the editor does. This should in theory fix the issue you having with the screen.

TitanNano:
@mrDIMAS thanks that does indeed work. For some reason the scaling of the inverse transform is not applied when I use the `transform_size` function, but when I use the inverse transform directly via `.transform_vector` it does.

One thing I noticed is that I have to set the `layout_transform` on the root node and not the `render_transform`. Setting the `layout_transform` to `2.0` turns a 50px margin into 100px, but doing the same for the `render_transform`, creates a 200px margin, so 4x scaling. This only happens inside the screen widget, the editor ui scales the same with the layout or the render transform.

Are you fine with keeping the automatic scaling of the root node?

mrDIMAS:
Right, `transform_size` should be used only if there's `NaN`s and `Inf`s in the vector, but screen size is always valid so `transform_vector` is working ok. As for global scaling code - this change is rather dangerous so I'd remove it.

PR #795, opened 2025-04-29
Improved focus distance to be less close to the object

adriengivry:
## Description

Improved editor camera focus distance to be less close to the object, effectively allowing the user to view the object they chose to focus.

## Related Issue(s)

_N/A_

## Review Guidance

_N/A_

## Screenshots/GIFs

This is what focusing looks like before and after the change
| Before Change | After Change |
| - | - |
| image | image |
| Way too close to the object to see anything (even closer than z_near) | Slightly further away |

## Checklist

### DIFF
--- a/editor/src/camera/mod.rs
+++ b/editor/src/camera/mod.rs
@@ -241,7 +241,7 @@ impl CameraController {
         world_space_position
     }
 
-    pub fn fit_object(&mut self, scene: &mut Scene, handle: Handle<Node>) {
+    pub fn fit_object(&mut self, scene: &mut Scene, handle: Handle<Node>, scale: Option<f32>) {
         // Combine AABBs from the descendants.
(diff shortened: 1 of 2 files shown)


editor/src/camera/mod.rs
```diff
@@ -256,6 +256,10 @@ impl CameraController {
             aabb = AxisAlignedBoundingBox::from_point(scene.graph[handle].global_position());
         }
 
+        if scale.is_some() {
```
mrDIMAS, line 259: `if let Some(scale) = scale { .. }`

mrDIMAS (changes requested):

editor/src/camera/mod.rs
adriengivry: Addressed by 8458b84c073cd75e5689c21b35d6d9d3cb7d4551

mrDIMAS (approved):

PR #814, opened 2025-06-01
Editor improvements

kkolyan:
## Description
some fixes and improvements for editor (see git comments)


mrDIMAS:
Could you please fix the project manager as well? It does not compile right now because of the additional arguments you've added to some methods.

kkolyan:
dropped one commit, because `Log::info("Dynamic plugins reloaded.");` caused log spam in project manager

mrDIMAS (approved):

PR #840, opened 2025-09-18
Refactored the Script API with more consistent naming and added fine-grained error handling

### LINKED ISSUE #837: Cannot obtain Handle<RigidBody> from &RigidBody programmatically.
## Description
I am trying to call Joint::set_body1() to set the rigid body of a joint programmatically. However, the signature of the function is:

Image

The type of handle is Handle\<RigidBody\>.
I couldn't find any other way to produce a Handle\<RigidBody\> from an existing RigidBody other than:

Image

which is quite annoying.

I saw in the Fyrox book that the previous version of Joint takes Handle\<Node\> as its bodies. I think this works fine and don't get why this is changed.

Image

Also, it seems Handle\<RigidBody\> is only used by the Joint and no other handle types like Handle\<Collider\>, Handle\<Joint\> ever appeared in the engine, so it is quite weird to have RigidBody as an exception.

## Proposal
Just to avoid confusion and let users know how to get what they want:

* Add comments to all the functions that take a Handle\<T\> other than Handle\<Node\> to explain how to convert one from Handle\<Node\>, or let these functions take a Handle\<Node\> directly.
* Add a section in the Fyrox book specializing in discussing how to convert among &Node, Handle\<Node\>, &RigidBody, Handle\<RigidBody\>, etc., under different contexts (ScriptContext, etc.)
* Add a section in the Fyrox book discussing about the hidden functionalities of object retrievers like context.scene.graph. For example, Graph::try_get_mut_of_type() function does not appear in the intelli-sense until you specify "use fyrox::graph::SceneGraph" manually.

Additionally, to promote the use of typed Handle for communication and maintenance purpose:

* Make the method handle() return Handle\<ActualType\> for types implementing NodeTrait. For example, rigidbody.handle() returns Handle\<RigidBody\>. Name the original handle() method that returns Handle\<Node\> to node_handle().
* Make Graph::get() and get_mut() method return the reference to the exact type T specified by the Handle\<T\>. For example, Handle\<Node\> -> &Node, Handle\<RigidBody\> -> &RigidBody. Rename methods that produce a reference of a different type to something else. For example, Graph::get_as_node(), Graph::get_as_type::\<T\>().
* Sync all the methods of Graph with its mbc (multiple borrow checker) version. Currently their methods' naming conventions are not united. Also consider moving hidden trait methods (like the current try_get_mut_of_type) to impl Graph{...} directly, so that beginners can pick their expected functionalities directly from the intelli-sense.
* Add a method Handle::try_into_handle_of_type::\<T\>() that converts a Handle directly to a Handle of another type. Return None if the handle is invalid or of a different type.
* Add a method Handle::is_correct_type() that returns a boolean indicating whether its stated type matches the object's type it points to.

LuoZheng2002:
## Description

Previously the Script API for context.scene.graph.* and related methods are confusing to me as a beginner, and critical functionalities are not automatically exposed to users. For example:
1. The immediately visible try_get and try_get_mut only returns Node, while methods that return the actual node variant like Rigidbody are hidden from the intelli-sense unless the corresponding traits (BaseSceneGraph, SceneGraph) are manually imported by the user (because these traits are in a different crate). 
2. The naming conventions are quite chaotic and there are many duplicated logic, which can easily confuse users who inspect into the code and try to figure out what these functions do. For example, the method that accepts both handle to Node and handle to a node variant (like Rigidbody) is called "typed_ref". Apart from its ambiguous name, its implementation involves "reverse of control" between pool and Handle through the BorrowAs trait. I get why this is necessary, but I figured out a more straightforward way to achieve the same effect.
3. The API for Graph and MultiBorrowContext is not consistent for commonly used functionalities, which does not look very professional.
4. The method that converts between Handle\<T\> and Handle\<U\> is called Handle::transmute, while the method that converts Node to its underlying variant is called Node::cast. Users may try to search for "cast" for the first case and get.../try... for the second case. These methods are essentially hidden from the users.

## Related Issue(s)

Fixes #837

## Review Guidance

The major work done is as follows:
1. Renamed typed_ref to try_get, since the most used functionality deserves the most vanilla name.
2. Renamed try_get to try_get_node, preserving its original semantics. For script APIs, try_get_node is hidden because the same functionality can be achieved with try_get.
3. Exposed previously hidden script APIs like try_get_of_type by adding "use fyrox::graph::prelude::*" in the script template.
4. Renamed Handle::transmute to Handle::cast, since "transmute" is hard to be bumped into by beginners, semantically inappropriate (according to ChatGPT) and verbose to be used extensively.
5. Removed get_* from script APIs to reduce API complexity. Users can use try_get_* + unwrap.
6. Unified the API for Graph and MultiBorrowContext, including try_get and try_get_component.
7. Unified error handling for Graph and MultiBorrowContext, and added MismatchedType variant for retrieving a specified node variant type. For NoSuchComponent variant, it now carries information about the actual type and visited component types.
8. Used a cleaner implementation for try_get with the introduced NodeOrNodeVariant\<T\> trait. The BorrowAs trait is deleted and no "reverse of control" is needed.
9. Pool and associated structures now only take 1 generic parameter instead of 2. NodeContainer and WidgetContainer is discarded. This is achieved by introducing the VisitAsOption trait, which bypasses the restriction that impl Visit for Option\<T\> requires T: Default in a safe way. The NodeContainer and WidgetContainer's Visit logic is transferred into the implementation of VisitAsOption for Node and UiNode. Then Payload\<P\> can be reduced to UnsafeCell\<Option\<NodeType\>\>, effectively eliminating a redundant generic parameter P.
10. The derive proc macro for ComponentProvider is revised to support richer diagnostic information on error.
11. Adapted rest of the code base to the API change (for example, replacing Some -> Ok, is_some -> is_ok on try_get* calls).

These changes are not fully documented yet. I'll finish the documentation after these ideas are accepted.

These changes are tested on a simple project. The editor and the game didn't crash. No unexpected behaviors. Existing auto tests are all passed.

## Screenshots/GIFs
Current Graph::try_get* APIs in script:
image
Current MultiBorrowContext::try_get* APIs in script:
image
Current node conversion graph:
image

It may be better for &RigidBody to convert to Handle\<RigidBody\> directly.

## Checklist

### DIFF
--- a/fyrox-core/src/pool/mod.rs
+++ b/fyrox-core/src/pool/mod.rs
@@ -1289,7 +1289,6 @@ impl<T: 'static> Pool<T> {
     /// generation of pool record at handle index (in other words it means that object
     /// at handle's index is different than the object was there before).
     #[inline]
-    #[must_use]
     pub fn try_get_node(&self, handle: Handle<T>) -> Result<&T, BorrowError> {
         let record = self.records_get(handle.index).ok_or_else(|| {
             BorrowError::new(
@@ -1315,7 +1314,6 @@ impl<T: 'static> Pool<T> {
     /// generation of pool record at handle index (in other words it means that object
     /// at handle's index is different than the object was there before).
     #[inline]
-    #[must_use]
     pub fn try_get_node_mut(&mut self, handle: Handle<T>) -> Result<&mut T, BorrowError> {
         // self.records_get_mut(handle.index).and_then(|r| {
         //     if r.generation == handle.generation {
--- a/fyrox-core/src/visitor/mod.rs
+++ b/fyrox-core/src/visitor/mod.rs
@@ -384,18 +384,18 @@ pub trait Visit {
     fn visit(&mut self, name: &str, visitor: &mut Visitor) -> VisitResult;
 }
 
-/// Unify Visit implementation for Option<T> for T implementing Default and not implementing Default.
+/// Unify `Visit` implementation for `Option<T>` where `T` may or may not implement `Default`.
 /// See [`VisitAsOption::visit_as_option`] for more info.
 pub trait VisitAsOption: Sized {
-    /// Unify Visit implementation for Option<T> for T implementing Default and not implementing Default.
+    /// Unify `Visit` implementation for `Option<T>` where `T` may or may not implement `Default`.
     ///
-    /// For T implementing Default, Visit trait is implemented for Option<T>, and VisitAsOption serves as a wrapper for Payload<T> to call.
+    /// For `T` implementing `Default`, Visit trait is implemented for `Option<T>`, and `VisitAsOption` serves as a wrapper for `Payload<T>` to call.
     ///
-    /// For T not implementing Default, like Box<_>, this trait enables users to implement Visit manually for Payload<T> to call.
+    /// For `T` not implementing `Default`, like `Box<_>`, this trait enables users to implement Visit manually for `Payload<T>` to call.
     ///
-    /// In this way, we can unify the Visit implementation for Payload<T>.
+    /// In this way, we can unify the Visit implementation for `Payload<T>`.
     ///
-    /// The reason why deriving Visit for Option<T> from T directly is not sufficient is that we need to avoid uninitialized (default) values to appear before the T::visit call.
+    /// The reason why deriving Visit for `Option<T>` from `T` directly is not sufficient is that we need to avoid uninitialized (default) values to appear before the `T::visit` call.
     fn visit_as_option(
         option_self: &mut Option<Self>,
         name: &str,
--- a/fyrox-impl/src/scene/graph/mod.rs
+++ b/fyrox-impl/src/scene/graph/mod.rs
@@ -124,7 +124,7 @@ impl GraphPerformanceStatistics {
     }
 }
 
-/// A helper type alias for node pool.
+// /// A helper type alias for node pool.
 // pub type Pool<Node> = Pool<Node, NodeContainer>;
 
 // impl<T: NodeTrait> BorrowAs<Node, NodeContainer> for Handle<T> {
@@ -495,7 +495,7 @@ impl Graph {
         self.root
     }
 
-    /// Tries to mutably borrow a node, returns Some(node) if the handle is valid, None - otherwise.
+    // /// Tries to mutably borrow a node, returns Some(node) if the handle is valid, None - otherwise.
     // #[inline]
     // pub fn try_get_node_mut(&mut self, handle: Handle<Node>) -> Option<&mut Node> {
     //     self.pool.try_get_node_mut(handle)
--- a/fyrox-core/src/pool/multiborrow.rs
+++ b/fyrox-core/src/pool/multiborrow.rs
@@ -265,14 +265,14 @@ impl<'a, T: 'static> MultiBorrowContext<'a, T> {
         })
     }
 
-    /// Tries to get a mutable reference to a pool element located at the given handle. The method could
-    /// fail in three main reasons:
-    ///
-    /// 1) A reference to an element is already taken - returning multiple mutable references to the
-    /// same element is forbidden by Rust safety rules.
-    /// 2) You're trying to get more references that the context could handle (there is not enough space
-    /// in the internal handles storage) - in this case you must increase `N`.
-    /// 3) A given handle is invalid.
+    // /// Tries to get a mutable reference to a pool element located at the given handle. The method could
+    // /// fail in three main reasons:
+    // ///
+    // /// 1) A reference to an element is already taken - returning multiple mutable references to the
+    // /// same element is forbidden by Rust safety rules.
+    // /// 2) You're trying to get more references that the context could handle (there is not enough space
+    // /// in the internal handles storage) - in this case you must increase `N`.
+    // /// 3) A given handle is invalid.
     // #[inline]
     // pub fn try_get_node<'b: 'a>(
     //     &'b self,
--- a/fyrox-graph/src/lib.rs
+++ b/fyrox-graph/src/lib.rs
@@ -881,13 +881,10 @@ pub trait SceneGraph: BaseSceneGraph {
 
     /// Tries to borrow a node and fetch its component of specified type.
     #[inline]
-    fn try_get_of_type<T: NodeVariant<Self::NodeType>>(
+    fn try_get_of_type<T: NodeVariant<Self::NodeType> + 'static>(
         &self,
         handle: Handle<Self::Node>,
-    ) -> Result<&T, BorrowError>
-    where
-        T: 'static,
-    {
+    ) -> Result<&T, BorrowError> {
         // self.try_get_node(handle)
         //     .and_then(|n| n.query_component_ref(TypeId::of::<T>()))
         //     .and_then(|c| c.downcast_ref())
@@ -896,13 +893,10 @@ pub trait SceneGraph: BaseSceneGraph {
 
     /// Tries to mutably borrow a node and fetch its component of specified type.
     #[inline]
-    fn try_get_mut_of_type<T: NodeVariant<Self::NodeType>>(
+    fn try_get_mut_of_type<T: NodeVariant<Self::NodeType> + 'static>(
         &mut self,
         handle: Handle<Self::Node>,
-    ) -> Result<&mut T, BorrowError>
-    where
-        T: 'static,
-    {
+    ) -> Result<&mut T, BorrowError> {
         // self.try_get_node_mut(handle)
         //     .and_then(|n| n.query_component_mut(TypeId::of::<T>()))
         //     .and_then(|c| c.downcast_mut())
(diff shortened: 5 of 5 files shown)


LuoZheng2002:
Thanks! If you have any questions about the implementations, feel free to ask.

fyrox-core/src/visitor/mod.rs
```diff
@@ -384,6 +384,25 @@ pub trait Visit {
     fn visit(&mut self, name: &str, visitor: &mut Visitor) -> VisitResult;
 }
 
+/// Unify `Visit` implementation for `Option<T>` where `T` may or may not implement `Default`.
+/// See [`VisitAsOption::visit_as_option`] for more info.
+pub trait VisitAsOption: Sized {
+    /// Unify `Visit` implementation for `Option<T>` where `T` may or may not implement `Default`.
+    ///
+    /// For `T` implementing `Default`, Visit trait is implemented for `Option<T>`, and `VisitAsOption` serves as a wrapper for `Payload<T>` to call.
```
mrDIMAS, line 392: Do mention unrelated types here - Payload is a part of Pool.

fyrox-impl/src/engine/hotreload.rs
```diff
@@ -89,10 +91,11 @@ impl SceneState {
                 // The entire node belongs to plugin, serialize it entirely.
                 // Take the node out of the graph first.
                 let (ticket, node) = scene.graph.take_reserve(handle);
-                let mut container = NodeContainer::new(node);
+                // Using Node::visit_as_option makes the guarantee that writing to the Node is safe (no uninitalized Node (Box dyn NodeTrait) ever appears)
```
mrDIMAS, line 94: Again, do not mention types from different crate here. The rule here is that you can't mention types of the crates down the hierarchy from the crates up in the hierarchy.

fyrox-impl/src/engine/hotreload.rs
```diff
@@ -225,9 +228,9 @@ impl SceneState {
                     widget_constructors,
                 )
                 .map_err(|e| e.to_string())?;
-                let mut container = NodeContainer::default();
-                container
-                    .visit("Node", &mut visitor)
+                // Use visit_as_option to make sure no uninitialized Node (Box<dyn NodeTrait>) ever appears
```
mrDIMAS, line 231: Same here.

fyrox-impl/src/scene/graph/mod.rs
```diff
@@ -123,30 +124,30 @@ impl GraphPerformanceStatistics {
     }
 }
 
-/// A helper type alias for node pool.
-pub type NodePool = Pool<Node, NodeContainer>;
+// /// A helper type alias for node pool.
```
mrDIMAS, line 127: ?

fyrox-impl/src/scene/graph/mod.rs
```diff
@@ -494,11 +495,11 @@ impl Graph {
         self.root
     }
 
-    /// Tries to mutably borrow a node, returns Some(node) if the handle is valid, None - otherwise.
-    #[inline]
-    pub fn try_get_mut(&mut self, handle: Handle<Node>) -> Option<&mut Node> {
-        self.pool.try_borrow_mut(handle)
-    }
+    // /// Tries to mutably borrow a node, returns Some(node) if the handle is valid, None - otherwise.
```
mrDIMAS, line 498: ?

fyrox-impl/src/scene/graph/mod.rs
```diff
@@ -1629,32 +1632,64 @@ impl Graph {
...
         self.instance_id_map
             .get(&id)
-            .and_then(|h| self.pool.try_borrow_mut(*h).map(|n| (*h, n)))
+            .and_then(|h| self.pool.try_get_node_mut(*h).ok().map(|n| (*h, n)))
     }
 }
 
-impl<T, B: BorrowAs<Node, NodeContainer, Target = T>> Index<B> for Graph {
-    type Output = T;
+// impl<T, B: BorrowAs<Node, NodeContainer, Target = T>> Index<B> for Graph {
```
mrDIMAS, line 1646: ?

fyrox-impl/src/scene/node/container.rs
```diff
@@ -20,103 +20,76 @@
...
-    }
-}
+//!
+
+// Since Pool<Node> uses NodeContainer exclusively,
+// we just pass the Visit implementation to the Node through VisitAsOption trait.
+// See impl VisitAsOption for Node in fyrpx-impl/src/scene/graph/mod.rs.
+// (not fyrox-graph/src/lib.rs, which is just for testing)
+
+// use crate::{
```
mrDIMAS, line 30: ?

fyrox-impl/src/scene/node/container.rs
```diff
@@ -20,103 +20,76 @@
...
+//         reflect::prelude::*,
+//         uuid::Uuid,
+//         visitor::{Visit, VisitResult, Visitor},
+//     },
+//     engine::SerializationContext,
+//     scene::node::Node,
+// };
+// use fyrox_core::visitor::error::VisitError;
+
+// /// A wrapper for node pool record that allows to define custom visit method to have full
```
mrDIMAS, line 41: ?

fyrox-core/src/pool/mod.rs
```diff
@@ -65,40 +67,216 @@ const INVALID_GENERATION: u32 = 0;
...
+pub enum BorrowErrorKind {
+    Empty,
+    InvalidHandleIndex,
+    InvalidHandleGeneration,
+    MismatchedType(MismatchedTypeError),
+    NoSuchComponent(QueryComponentError),
+}
+
+#[derive(PartialEq)]
+pub struct BorrowError {
```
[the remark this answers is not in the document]
b-guild: We could also save those bytes by doing
```rust
pub struct BorrowError(Box<BorrowErrorInfo>);
pub struct BorrowErrorInfo {
  pub kind: BorrowErrorKind,
  pub handle_info: HandleInfo,
}
```
That way we do not have to give up the error reporting.

editor/src/plugins/probe.rs
```diff
@@ -191,6 +192,8 @@ impl InteractionMode for ReflectionProbeInteractionMode {
                 settings: &settings.selection,
             },
         ) {
+            fn is_static<T: 'static>() {}
```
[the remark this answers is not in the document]
LuoZheng2002: Oh, sorry about that. It's just for me debugging the trait bound errors. I'll remove it.

fyrox-core/src/pool/handle.rs
```diff
@@ -271,6 +271,11 @@ impl<T> Handle<T> {
         }
     }
 
+    #[deprecated(
+        note = "The term 'transmute' is not appropriate for describing the conversion between different handle types
+     since 'transmute' in Rust refers to a more dangerous type casting that interprets the underlying fields differently,
+     while in the case of handle only the nominal generic type is changed. 'cast' is a more suitable term according to ChatGPT."
```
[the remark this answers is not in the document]
LuoZheng2002: Sure.

fyrox-core/src/pool/mod.rs
```diff
@@ -65,40 +67,216 @@ const INVALID_GENERATION: u32 = 0;
...
-    fn borrow_as_mut(self, pool: &mut Pool<Object, Container>) -> Option<&mut Object> {
-        pool.try_borrow_mut(self)
+impl Debug for BorrowError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(self, f)
     }
 }
 
-impl<T, P> Reflect for Pool<T, P>
+// pub trait BorrowAs<Object> {
```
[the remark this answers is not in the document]
LuoZheng2002: I'll delete it. Its functionality is replaced by NodeOrNodeVariant trait.

fyrox-core/src/pool/mod.rs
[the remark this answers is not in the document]
LuoZheng2002: I agree. Maybe we need another version of try_get* that does not run error reporting. My initial thought was that failing to retrieve a reference is rare and undesirable, especially in the real game, in which case the performance different is negligible, but now I find that in the game engine many uses of them work like an if statement.
I think the best solution is to split the functionality into two classes. One returns Option and the other returns Error, so that we can pick the one best for the context. I can work on it if you would like me to.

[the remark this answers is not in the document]
b-guild: The issue isn't *rarity* since `Result<V,E>` always has the size of whichever of `V` or `E` is bigger, so you pay the price of a big `E` even when the result is `Ok`. Even if `try_get` never failed, we still pay the price for the size of `E`. That is why I suggest using a `Box` so that `E` will just be the size of a pointer.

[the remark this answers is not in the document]
mrDIMAS: It is definitely worth wrapping the error in Box. However, it could be slow if used for checking only (like `try_get(..).is_ok()`) so in this case a special, more lightweight, method is needed.

PR #846, opened 2025-09-29
Renamed pool method typed_ref to try_get

LuoZheng2002:
## Description

We need a uniform naming convention for the functionality of retrieving pool objects or their variants from the pool and the graph. Currently there are `typed_ref`, `try_get_of_type`, etc. I think it is beneficial to unify them into the vanilla name `try_get`. For current occurrences of `try_get` for the Graph, I renamed it to `try_get_node`. Specifying "node" for Graph is appropriate because `Graph` uses `Node` as the pool object type.

I also added a prelude module for the fyrox_graph crate containing `AbstractSceneGraph`, `BaseSceneGraph` and `SceneGraph`, and added `use fyrox::graph::prelude::*` in the script template so that functionalities like try_get are exposed by default.

## Review Guidance

The code passed all auto tests.

## Checklist

### DIFF
--- a/editor/src/asset/mod.rs
+++ b/editor/src/asset/mod.rs
@@ -837,7 +837,7 @@ impl AssetBrowser {
                 if let Some(item) = engine
                     .user_interfaces
                     .first_mut()
-                    .try_get(self.context_menu.placement_target)
+                    .try_get_node(self.context_menu.placement_target)
                     .and_then(|n| n.cast::<AssetItem>())
                 {
                     if let Ok(resource) =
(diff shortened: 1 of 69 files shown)


mrDIMAS:
Could you please fix the code generated by the project template generator? Check the CI for project template, it highlighted that there's an unused import there. The changes should be done in `template-core`

LuoZheng2002:
I've added the `#[allow(unused_imports)]` to the template. My idea is that using try_get methods happen in almost every script, but try_get cannot be exposed unless the users import the SceneGraph trait themselves. They would have to remember the exact directory `use fyrox::graph::prelude::*` because there is no one-click solution for that. (because traits like SceneGraph are defined in fyrox-graph while the Graph is defined in fyrox-impl)

mrDIMAS (approved):

PR #911, opened 2026-04-12
Add editor font size settings

### LINKED ISSUE #910: Add UI font size configuration in Editor Settings Window
## Problem this feature should fix
Add option for users to configure font sizes in Editor UI.

## Expected solution
Users should be able to manually set Editor UI font sizes from Settings Window.

CopeFiend:
Closes #910

## Description

Add options for configuring font size in Editor UI.

## Progress
- [x] Add `font_size` option to Settings Window -> General
- [x] Modify Editor font size through `font_size` option
- [x] Store modified value in `settings.ron`
- [x] Clamp value in Settings Window (8.0px - 48.0px)
- [ ] Handle text widgets with previously hard-coded values (World Viewer, Tilemap Property, etc.) 
- [ ] Handle layout scaling with large fonts

## Review Guidance

 - Most `Builders` now have the following fields and methods:
 ```rust
pub font: Option<FontResource>, // Default: None
pub font_size: Option<StyledProperty<f32>>, // Default: None

pub fn with_font(mut self, font: FontResource) -> Self {
    self.font = Some(font);
    self
}

pub fn with_font_size(mut self, font_size: StyledProperty<f32>) -> Self {
    self.font_size = Some(font_size);
    self
}
``` 
 - Every instance of `Builder` now explicitly calls the `with_font()` and `with_font_size()` methods
 - Would like to whether putting everything in Settings::general is a wise decision.

## Screenshots/GIFs

WIP

## Checklist

- [ ] My code follows the project's code style guidelines
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the documentation accordingly
- [ ] My changes don't generate new warnings or errors
- [ ] No unsafe code introduced (or if introduced, thoroughly justified and documented)


mrDIMAS:
I'm not quite sure why there's a need to add this huge heap of additional methods. What I'd do is to just modify standard [style's variable](https://github.com/FyroxEngine/Fyrox/blob/master/fyrox-ui/src/style/mod.rs#L348) and called `ui.set_style`. This will propagate all the changes to all widgets that handles text. Also, for now the editor uses this `FONT_SIZE` variable almost everywhere, so it is basically uses the same font size for every widget. 

For places that has different font size, I'd just add a separate style variable, specific to the place where it belongs. Then it can be modified and `ui.set_style` can be called. 

To proportionally change the font size for every styled variable I'd change the type of `FONT_SIZE` variable to a new type called `FontSize` and after that it'd be possible to iterate over all such properties in the style and scale them accordingly. So basically: take the default style, create its copy, scale all the `FontSize` properties accordingly, then call `ui.set_style`. This should be enough.

Of course, I might be missing some issues. Let's discuss this.

CopeFiend:
@mrDIMAS Created a Discussion thread #915 , with replies to feedback https://github.com/FyroxEngine/Fyrox/discussions/915#discussioncomment-16770660 .

PR #912, opened 2026-04-16
Fix relative path resolution for symlinked asset directories

### LINKED ISSUE #907: Can't save UI scene with assets stored in a symlink dir
## Description
My assets are in data, that is a symlink dir with assets. When I try to save a UI scene I get an error in log
```[ERROR]: Operation failed! Reason: No watch was found. about ["/home/user/data"]
[ERROR]: Failed to create relative path for /home/user/data/unnamed.ui. Reason: unable to strip prefix from '/home/user/data/unnamed.ui'! Reason: prefix not found
``` 
And saving doesn't work.
## To Reproduce
1. Create a symlink dir with assets
2. Save a UI scene
3.  Get errors, file can't be saved, can't change output directory

## Expected behavior
Scene is saved.

AghastyGD:
## Description

`make_relative_path` fails when asset directories are symlinks pointing outside the project. `canonicalize()` resolves symlinks to their real filesystem paths, so `strip_prefix` can no longer match the working directory and scene saving breaks entirely.

This PR adds a 3-step fallback strategy to `make_relative_path` in fyrox-core:

1. **Without canonicalization**: try `strip_prefix` on the raw path first, preserving symlinks. This handles the common case where the path already goes through the symlink textually.
2. **With canonicalization**: resolve `.` and `..` components and retry. This is the original behavior for non-symlink cases.
3. **Reverse symlink lookup**: if both fail, scan the working directory tree (up to 4 levels deep) for symlinks whose canonical target is a prefix of the resolved path, then reconstruct the relative path through that symlink.

A helper function `find_symlink_relative_path` was extracted for step 3 to keep things readable.

## Related Issue(s)

Fixes #907

## Review Guidance

The core change is in `fyrox-core/src/lib.rs`, function `make_relative_path`. No other files were modified.

The symlink scan (step 3) only runs as a last resort. Normal paths return at step 1 or 2 without any extra filesystem access. The recursive walk is capped at `MAX_DEPTH = 4` to avoid expensive traversals.

The test `test_make_relative_path_with_symlinks` covers both top-level symlinks (`project/data -> real_assets/`) and nested symlinks (`project/assets/textures -> real_textures/`), using both resolved and unresolved input paths. It is `#[cfg(unix)]` since symlinks behave differently on Windows. Both symlink scenarios are in a single test to avoid race conditions from the process-global `set_current_dir`.

## Screenshots/GIFs

N/A, this is a logic fix with no UI changes.

## Checklist

### DIFF
--- a/fyrox-core/src/lib.rs
+++ b/fyrox-core/src/lib.rs
@@ -329,6 +329,23 @@ pub fn make_relative_path<P: AsRef<Path>>(path: P) -> Result<PathBuf, std::io::E
     } else {
         Path::new(".")
     };
+
+    let cwd = std::env::current_dir()?;
+
+    // Try without canonicalization first to preserve symlinks.
+    let non_canon_dir = if dir.is_absolute() {
+        dir.to_path_buf()
+    } else {
+        cwd.join(dir)
+    };
+    if non_canon_dir.is_dir() {
+        let non_canon_path = non_canon_dir.join(file_name);
+        if let Ok(relative_path) = non_canon_path.strip_prefix(&cwd) {
+            return Ok(replace_slashes(relative_path));
+        }
+    }
+
+    // Canonicalize to resolve `.` and `..` components.
     let canon_path = dir
         .canonicalize()
         .map_err(|err| {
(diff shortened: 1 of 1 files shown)


AghastyGD:
Hi @mrDIMAS, I noticed the CI failed on `test_close_with_save`, so I took a closer look

It seems the failure is caused by parallel tests modifying the process-wide working directory (`set_current_dir`). When running with multiple threads, the working directory gets changed by other tests, leading to inconsistent paths (e.g. nested test directories).

Running the test in isolation passes consistently, which suggests this is a pre-existing race condition rather than an issue introduced by this change.

Happy to investigate further or adjust if needed.

mrDIMAS (approved):

mrDIMAS:
Oh wow, you've found the issue with these parallel editor tests that I couldn't find in a few attempts, very cool! I'll try to modify the testing harness to prevent such race condition.

As for the PR, looks good, merging.

AghastyGD:
> Oh wow, you've found the issue with these parallel editor tests that I couldn't find in a few attempts, very cool! I'll try to modify the testing harness to prevent such race condition.
> 
> As for the PR, looks good, merging.

Thanks a lot for the review and kind words, really appreciate it 🙂

PR #917, opened 2026-05-09
Fix transmute_slice UB: add alignment, divisibility, and ZST guards

### LINKED ISSUE #877: [Undefined Behavior] Misaligned Raw Pointer with `transmute_slice`
Hello, we are reviewing the previous advisories. We find that the following function `transmute_slice` can trigger undefined behavior by creating misaligned pointer. We consider that we should not allow users to decide two generic types (`v` and `U`) by themselves.

https://github.com/FyroxEngine/Fyrox/blob/ebffbe610b28fb761c083a6867480d4fc49969f0/fyrox-core/src/lib.rs#L366-L369

Check the safety [doc](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html).
> data must be non-null, valid for reads for len * size_of::<T>() many bytes, and it must be properly aligned.

Following is the PoC:
```rust
use fyrox_core::transmute_slice;

fn main() {
    let data: Vec<u8> = vec![1, 2, 3, 4, 5];
    let bytes: &[u32] = transmute_slice(&data);
    println!("Byte length: {}", bytes.len());
}
```
Miri's result:
```sh
error: Undefined Behavior: constructing invalid value: encountered an unaligned reference (required 4 byte alignment but found 1)
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fyrox-core-0.36.2/src/lib.rs:338:9
    |
338 | /         std::slice::from_raw_parts(
339 | |             v.as_ptr() as *const U,
340 | |             std::mem::size_of_val(v) / std::mem::size_of::<U>(),
341 | |         )
    | |_________^ Undefined Behavior occurred here
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE:
    = note: inside `fyrox_core::transmute_slice::<u8, u32>` at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fyrox-core-0.36.2/src/lib.rs:338:9: 341:10
```
We suggest to add assertion of alignment check.

kimjune01:
## Summary

- Adds runtime alignment assertion before `slice::from_raw_parts` to prevent misaligned pointer UB (reported in #877)
- Adds byte-length divisibility assertion to catch truncated trailing bytes
- Adds ZST guard to prevent division by zero
- Adds `Pod` trait bounds to `fyrox-core::transmute_slice` to ensure valid bit patterns for any byte sequence
- Applies same alignment + divisibility guards to `fyrox-texture`'s private `transmute_slice` and `transmute_slice_mut`

## Test plan

- [x] `cargo check -p fyrox-core -p fyrox-texture -p fyrox-graphics` passes
- [x] Existing tests in `fyrox-graphics/src/uniform.rs` that call `transmute_slice` still pass
- [ ] CI should verify no regressions across the full test suite

Closes #877

### DIFF
--- a/fyrox-core/src/lib.rs
+++ b/fyrox-core/src/lib.rs
@@ -413,13 +413,34 @@ pub fn array_as_u8_slice_mut<T: Sized + Pod>(v: &mut [T]) -> &'_ mut [u8] {
 }
 
 /// "Transmutes" array of any sized type to a slice of some other type.
-pub fn transmute_slice<T: Sized, U: Sized>(v: &[T]) -> &'_ [U] {
-    // SAFETY: It is safe to reinterpret data to read it.
+///
+/// Both `T` and `U` must be [`Pod`] types (no padding, no invalid bit patterns).
+///
+/// # Panics
+///
+/// Panics if:
+/// - `U` is a zero-sized type
+/// - The byte length of `v` is not a multiple of `size_of::<U>()`
+/// - The pointer to `v` is not properly aligned for type `U`
+pub fn transmute_slice<T: Sized + Pod, U: Sized + Pod>(v: &[T]) -> &'_ [U] {
+    let byte_len = std::mem::size_of_val(v);
+    let target_size = std::mem::size_of::<U>();
+    assert!(target_size != 0, "transmute_slice: target type must not be zero-sized");
+    assert!(
+        byte_len % target_size == 0,
+        "transmute_slice: source byte length ({byte_len}) is not a multiple of target type size ({target_size})",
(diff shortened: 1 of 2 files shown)


mrDIMAS:
AI-assisted code is not allowed due to the rules here https://github.com/FyroxEngine/Fyrox/blob/master/CONTRIBUTING.md#contributing-code

PR #918, opened 2026-05-09
Fix read_pixels_of_type UB: use bytemuck::cast_slice

### LINKED ISSUE #827: `GpuFrameBufferTrait::read_pixels_of_type` contains UB
I was reading the documentation of `std::mem::forget`, and under the "Relationship with ManuallyDrop" section it presents the following example: 
```rust
use std::mem;

let mut v = vec![65, 122];
// Build a `String` using the contents of `v`
let s = unsafe { String::from_raw_parts(v.as_mut_ptr(), v.len(), v.capacity()) };
// leak `v` because its memory is now managed by `s`
mem::forget(v);  // ERROR - v is invalid and must not be passed to a function
assert_eq!(s, "Az");
// `s` is implicitly dropped and its memory deallocated.
```
Notice that is almost exactly the same as the implementation of the ``GpuFrameBufferTrait::read_pixels_of_type` function. Then states that:

"After calling v.as_mut_ptr() and transmitting the ownership of the data to s, the v value is invalid. Even when a value is just moved to mem::forget (which won't inspect it), some types have strict requirements on their values that make them invalid when dangling or no longer owned. Using invalid values in any way, including passing them to or returning them from functions, constitutes undefined behavior and may break the assumptions made by the compiler".

And then provides a non-UB example using `ManuallyDrop`.

I'm not very experienced with Rust so Idk if this is really the case, so tell me if this should be fixed and I'll make a PR.

kimjune01:
## Summary

- Replaces unsafe `Vec::from_raw_parts` + `mem::forget` with safe `bytemuck::cast_slice` + `to_vec()`
- The previous implementation had two sources of UB:
  1. `mem::forget` received a `Vec<u8>` whose memory was already transferred to another `Vec`
  2. `Vec<T>` would deallocate with `align_of::<T>()` but the memory was allocated with `align_of::<u8>() = 1`
- The new implementation safely casts the byte slice and copies into a properly-aligned `Vec<T>`
- `bytemuck` is already a dependency of `fyrox-graphics`

## Test plan

- [x] `cargo check -p fyrox-graphics` passes
- [ ] CI should verify no regressions across the full test suite

Closes #827

### DIFF
--- a/fyrox-graphics/src/framebuffer.rs
+++ b/fyrox-graphics/src/framebuffer.rs
@@ -328,16 +328,11 @@ impl dyn GpuFrameBufferTrait {
     where
         T: Pod,
     {
-        let mut bytes = self.read_pixels(read_target)?;
-        let typed = unsafe {
-            Vec::<T>::from_raw_parts(
-                bytes.as_mut_ptr() as *mut T,
-                bytes.len() / size_of::<T>(),
-                bytes.capacity() / size_of::<T>(),
-            )
-        };
-        std::mem::forget(bytes);
-        Some(typed)
+        let bytes = self.read_pixels(read_target)?;
+        // Use bytemuck for a safe, correctly-aligned conversion. This copies into a new
+        // Vec<T> with proper alignment, avoiding the UB from reinterpreting a Vec<u8>
+        // allocation as Vec<T> (which would deallocate with the wrong layout).
+        Some(bytemuck::cast_slice::<u8, T>(&bytes).to_vec())
     }
 }
 
--- a/fyrox-graphics/src/framebuffer.rs
+++ b/fyrox-graphics/src/framebuffer.rs
@@ -322,17 +322,63 @@ pub trait GpuFrameBufferTrait: GpuFrameBufferAsAny {
     ) -> Result<DrawCallStatistics, FrameworkError>;
 }
 
+/// Owned, typed view of pixels read back from a [`GpuFrameBufferTrait`]. Preserves the
+/// original byte allocation (and therefore its memory layout) so it is dropped with the
+/// same `Layout` it was allocated with, while exposing a `&[T]` view of the contents.
+///
+/// This avoids the UB of reinterpreting a `Vec<u8>` allocation directly as `Vec<T>`
+/// (which would deallocate with `Layout::array::<T>(cap)` instead of the original
+/// `Layout::array::<u8>(cap)`) without copying the data.
+pub struct TypedPixels<T: Pod> {
+    bytes: Vec<u8>,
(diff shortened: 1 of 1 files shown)


fyrox-graphics/src/framebuffer.rs
```diff
@@ -328,16 +328,11 @@ impl dyn GpuFrameBufferTrait {
...
-                bytes.capacity() / size_of::<T>(),
-            )
-        };
-        std::mem::forget(bytes);
-        Some(typed)
+        let bytes = self.read_pixels(read_target)?;
+        // Use bytemuck for a safe, correctly-aligned conversion. This copies into a new
+        // Vec<T> with proper alignment, avoiding the UB from reinterpreting a Vec<u8>
+        // allocation as Vec<T> (which would deallocate with the wrong layout).
+        Some(bytemuck::cast_slice::<u8, T>(&bytes).to_vec())
```
mrDIMAS: `.to_vec` call is kinda concerning here, I'm pretty sure that this can be done without calling it. Probably by creating a proxy struct that keeps the memory layout like in here - https://github.com/FyroxEngine/Fyrox/blob/master/fyrox-impl/src/scene/mesh/buffer.rs#L269-L277

mrDIMAS (changes requested):

kimjune01:
@mrDIMAS thanks for the pointer to `BytesStorage` — addressed in 1d26960.

Introduced a `TypedPixels<T: Pod>` proxy struct that owns the original `Vec<u8>` and exposes `&[T]` via `Deref` (using `bytemuck::cast_slice`). The byte buffer is kept intact, so it's freed with the same `Layout::array::<u8>(cap)` it was allocated with. No `.to_vec()`, no copy.

I didn't carry an explicit `Layout` field like `BytesStorage` does, because that field is only needed when `BytesStorage` later decomposes back via `from_raw_parts` and deallocates manually. Here the `Vec<u8>` is never decomposed — dropping it as `Vec<u8>` already uses the right layout. Happy to add the field for symmetry if you'd prefer.

The single caller (`fyrox-impl/src/renderer/hdr/mod.rs:207`, `evaluator.average_luminance(&pixels)`) compiles unchanged via deref coercion. `cargo check -p fyrox-graphics` and `cargo check -p fyrox-impl` both pass.

mrDIMAS (approved):

du82:
This "Kim June" guy is using AI to generate and submit pull requests. I suggest blocking him.

mrDIMAS:
Yeah, but these changes were super obvious and I'm kinda ok with that. Anything more complex might not pass the review, because it would require real knowledge :)

Issue #721, opened 2025-02-05
Add `wgpu` graphics server


mrDIMAS:
Fyrox now supports custom graphics servers, which essentially is a tiny abstraction layer from graphics API. Currently there's only one graphics server, based on OpenGL 3.3/OpenGL ES 3.0. `wgpu` looks like a very good cross-platform replacement for OpenGL and it is definitely worth implementing a graphics server with it.

nice0hack:
Since no one has yet submitted a PR for wgpu integration, I'll continue the issue. The integration itself is very complex due to fundamental differences between OpenGL and wgpu concepts.

There's already a PR where you can run the engine on a raw wgpu implementation: https://github.com/FyroxEngine/Fyrox/pull/927

The implementation is raw, but the wgpu engine is ready for a first look.

This issue definitely needs to be addressed by the entire Fyrox Engine community, and the engine should gradually be fully migrated to wgpu. My tests show that wgpu is indeed a good replacement for OpenGL.

PR #927, opened 2026-07-26
Add wgpu graphics server

nice0hack:
## Description

This PR introduces the initial draft implementation of the `wgpu` backend for the Fyrox Engine.

A critical aspect of this PR is that **backward compatibility with the existing OpenGL backend is fully preserved**. The new `wgpu` integration is wrapped as an optional Cargo feature, meaning it does not affect the current default engine behavior.
To test the engine with the new backend, you need to run it using the following flags:
`cargo run --package fyroxed --profile=editor-standalone --no-default-features --features backend_wgpu`

One of the greatest advantages of `wgpu` introduced in this PR is its smart backend selection. It dynamically evaluates the most relevant graphics APIs available on the user's OS (e.g., DirectX 12, Vulkan, Metal) and automatically selects the most performant one. For example, on my Windows 10 machine with an older GPU, it seamlessly fell back to Vulkan to ensure optimal compatibility and performance.

## Related Issue(s)

https://github.com/FyroxEngine/Fyrox/issues/721

## Review Guidance

**Testing Environment:**
So far, this implementation has only been tested on **Windows 10** and **Fedora Linux 43**. Feedback from macOS or other environments would be highly appreciated.

**Known Bugs & Limitations (To-Do):**
I have encountered a few complex bugs that I haven't been able to resolve yet. Fixing them might require a significant amount of time and deep debugging:

* [ ] A mesh object in deferred rendering moves away from its original position depending on the camera's rotation.
* [ ] When adding any external light source, the entire scene rendering breaks.
* [ ] Debug rendering doesn't render properly.
* [ ] The camera in orthographic projection breaks completely.

**Discussion: MSAA & Deferred Rendering:**
Currently, MSAA (Multisample Anti-Aliasing) support is not implemented for the `wgpu` backend. Integrating MSAA with `wgpu` in our Deferred Rendering pipeline would cause a massive drop in performance and heavily inflate VRAM usage. We should discuss whether it's worth implementing MSAA at all, or if we should rely on post-processing anti-aliasing solutions (like FXAA or SMAA), which fit the deferred pipeline much better.

**Final Thoughts & Proposal for Merging:**
Integrating `wgpu` is a massive and fundamental shift for the engine's architecture. Since my changes are entirely isolated behind a feature flag (`backend_wgpu`) and do not break or affect the stable OpenGL backend, **I highly recommend merging this PR in its current state.**

Having this foundational codebase in the `master` branch will allow the community to test it, try out the `wgpu` backend, and contribute. We can then split the remaining bugs listed above into separate, manageable issues and gradually polish the `wgpu` transition together rather than keeping this giant update stuck in a single PR.

## Checklist

* [x] My code follows the project's code style guidelines
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have updated the documentation accordingly
* [x] My changes don't generate new warnings or errors
* [x] No unsafe code introduced (or if introduced, thoroughly justified and documented)


fyrox-graphics-wgpu/src/framebuffer.rs
```diff
@@ -0,0 +1,1134 @@
...
+    }
+}
+
+/// Expected vertex attribute locations that the standard shaders may need but
+/// geometry might not provide (e.g. boneWeights, boneIndices, vertexSecondTexCoord).
+///
+/// Each entry is a `(location, format, &'static [VertexAttribute])` triple. The
+/// attribute arrays are `const` statics so they have `'static` lifetime without
+/// needing `Box::leak`, avoiding per-draw-call memory leaks.
+const EXTRA_VERTEX_LAYOUTS: &[(u32, &[wgpu::VertexAttribute])] = &[
```
mrDIMAS, line 959: This looks like a hack, because it adds info about vertex layout in frame buffer itself. It is not necessarily a bad thing for a prototype, but looks like something that needs to be addressed later.

mrDIMAS:
Finally finished the review, everything's fine. That one commented issue can be fixed later.

mrDIMAS:
WASM CI still failing - `error[E0433]: cannot find module or crate `fyrox_graphics_wgpu` in this scope`

